PR #12, opened 2023-09-18
fix tag:Session issue

knavapbc:
PLT-140


gsf (approved):
This should fix that issue with the instance profile role assuming these roles.

PR #27, opened 2023-12-18
PLT-176 Switch opt-out-inbound workflow from inbound to import

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-176

## 🛠 Changes

Change references from opt-out-inbound to opt-out-import.

## ℹ️ Context for reviewers

Standardizing the two sides of the opt-out process on import and export.

## ✅ Acceptance Validation

This will need to be merged before we can test the workflow, and checks will fail for now because they depend on code in #11.

## 🔒 Security Implications

None.


gfreeman-navapbc (approved):
LGTM

PR #31, opened 2023-12-22
Volume encryption automation

devopsmatt:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-220

## 🛠 Changes

The included script creates an encrypted copy of the AMI passed to it as parameter.

## ℹ️ Context for reviewers

This work is meant to help reduce/eliminate instances with unencrypted root volumes. 

## ✅ Acceptance Validation

A new Jenkins AMI with encrypted root filesystem has been created using the provided script -`ami-076e72264a6f55452`

## 🔒 Security Implications

None.


devopsmatt:
Corresponding PRs:
https://github.com/CMSgov/dpc-ops/pull/642
https://github.com/CMSgov/bcda-ops/pull/1014

.gitignore
```diff
@@ -7,3 +7,6 @@ node_modules
 terraform.tfstate*
 terraform.tfvars
 .terraform
+
+# macos cruft
+.DS_Store
```
gsf, line 12: This should be set in one's global gitignore.

scripts/encrypt_ami.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+        new_ami_id = response['ImageId']
+        print("New Encrypted AMI ID:", new_ami_id)
+
+        # Assign tags to the new AMI
+        ec2.create_tags(Resources=[new_ami_id], Tags=original_tags)
+
+    except Exception as e:
+        print(f"An error occurred: {e}")
```
gsf, line 43: No need for this try/except wrapper.

gsf (commented):
I'm a bit confused by the need for this script. If we're producing an unencrypted AMI, shouldn't we update the packer scripts responsible for that AMI to ensure it's encrypted?

.gitignore
devopsmatt: Good point.  I'll remove from here.
@christopher-maboh for visibility.

devopsmatt: Packer already takes AMI as [user param](https://github.com/CMSgov/bcda-ops/blob/master/packer/jenkins.json#L14), so as long as it is passed the AMI ID ami-076e72264a6f55452 or other encrypted AMI base, it. should work without changes.  I'll try to find appropriate documentation and update to reflect this.

scripts/encrypt_ami.py
devopsmatt: I'll update.
@christopher-maboh FYI

.gitignore
gsf: @christopher-maboh These lines can still be removed.

scripts/encrypt_ami.py
```diff
@@ -0,0 +1,50 @@
...
+    Create an encrypted copy of an AMI, copy its tags, add a new tag, and print the new AMI ID.
+
+    :param ami_id: The ID of the source AMI.
+    :param optional region: The region where the AMI is located. Default is 'us-east-1'.
+    :param optional description: A description for the new AMI.
+    """
+    # Create an EC2 resource session
+    ec2 = boto3.client('ec2', region_name=region)
+
+    try:
```
gsf, line 18: The `try:` and the extra level of indentation through the rest of this function also need to be removed.

gsf (approved):
A shared requirements.txt for scripts would get messy fast, so I've moved the encrypt_ami.py script into its own folder.

PR #34, opened 2023-12-29
PLT-278 Fix GitHub runner policy

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-278

## 🛠 Changes

Modified path and resource to fix IAM policy for GitHub runner.

## ℹ️ Context for reviewers

The permissions boundary for the deploy role won't allow management of policies outside of the /delegatedadmin/developer/ path, and I updated the deploy roles to include the application environment.

## ✅ Acceptance Validation

I had to delete the policy manually with an IAM user with direct access to the account. This should now apply cleanly once merged.

## 🔒 Security Implications

None.


gfreeman-navapbc (approved):
LGTM

PR #35, opened 2023-12-29
PLT-277 Switch gold image vars to aws params

gsf:
## 🎫 Ticket

## 🛠 Changes

Switched out gold image AMI variables to pull in values from AWS Parameter Store. Also updated deploy role.

## ℹ️ Context for reviewers

In an effort to keep secrets and variables in AWS instead of GitHub, I've set gold image params in the management account.

## ✅ Acceptance Validation

See checks. The build is failing due to an IAM PassRole issue with the bcda-packer role, likely because it isn't within the /delegatedadmin/developer/ path. We should still merge this and address that in another PR.

## 🔒 Security Implications

None.

### DIFF
--- a/.github/workflows/build-runner-images.yml
+++ b/.github/workflows/build-runner-images.yml
@@ -8,7 +8,7 @@ on:
     # 00:00 on Monday each week
     - cron: "0 0 * * 1"
   workflow_dispatch:
-    
+
 jobs:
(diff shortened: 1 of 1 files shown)


gfreeman-navapbc (approved):
LGTM, error in build to be addressed in follow up PR

PR #37, opened 2024-01-03
PLT-285: Image building must not use bcda-packer role

gfreeman-navapbc:
## 🎫 Ticket

## 🛠 Changes

Changes the role for building the images. The bcda-packer role is not within the permissions boundary.

## 🔒 Security Implications

None.


gsf (approved):
Looks good!

PR #38, opened 2024-01-05
PLT-278 Allow runner to get parameters

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-278

## 🛠 Changes

Allow the GitHub Actions runner role to retrieve parameters from the management account.

## ℹ️ Context for reviewers

This allows for workflows like https://github.com/CMSgov/ab2d/blob/7bfa9085ade3537f5a2dd4203debd936aa1b4814/.github/workflows/unit-integration-test.yml to set shared ARTIFACTORY and SONAR params from the management account before assuming a role.

## ✅ Acceptance Validation

See checks for terraform plan. I'll test with the workflow above after merge.

## 🔒 Security Implications

None.

### DIFF
--- a/terraform/services/github-actions-runner/main.tf
+++ b/terraform/services/github-actions-runner/main.tf
@@ -37,7 +37,7 @@ resource "aws_iam_policy" "runner" {
   name = "github-actions-runner"
   path = "/delegatedadmin/developer/"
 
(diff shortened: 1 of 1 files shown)


gsf:
Merge failed due to boundary policy. I've successfully applied locally.

PR #40, opened 2024-01-05
PLT-285 Add instance profile resource in deploy role terraform

gsf:
## 🎫 Ticket

## 🛠 Changes

Added instance profile for the deploy role.

## ℹ️ Context for reviewers

Workflow runs for #37 are failing because there is no instance profile for the deploy role.

## ✅ Acceptance Validation

We'll need to test in #37 following apply.

## 🔒 Security Implications

None.


gfreeman-navapbc (approved):
LGTM

PR #41, opened 2024-01-06
PLT-285 Drop instance profile from packer sources

gsf:
## 🎫 Ticket

## 🛠 Changes

Dropped iam_instance_profile from packer sources.

## ℹ️ Context for reviewers

It appears that iam_instance_profile is only needed when using session manager to connect to the instance for building. 

https://github.com/hashicorp/packer-plugin-amazon/blob/2f6ed10096b416c957b7a33d198d19f5e6000971/docs-partials/builders/aws-session-manager.mdx#L10

## ✅ Acceptance Validation

See checks.

## 🔒 Security Implications

None.

### DIFF
--- a/.github/workflows/build-runner-images.yml
+++ b/.github/workflows/build-runner-images.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     paths:
       - .github/workflows/build-runner-images.yml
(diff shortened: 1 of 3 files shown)


gfreeman-navapbc (approved):
LGTM

PR #43, opened 2024-01-10
PLT-277: Switch GitHub runners to use the AMI we build

gfreeman-navapbc:
## 🎫 Ticket

## 🛠 Changes

Changing the image our runner image builds work on to the latest prebuilt runner image.

## ℹ️ Context for reviewers

We successfully built an image using packer, but we are still using the old amzn2legacy image.

## ✅ Acceptance Validation

Image builds successfully.

## 🔒 Security Implications

- [ ] This PR adds a new software dependency or dependencies.
- [ ] This PR modifies or invalidates one or more of our security controls.
- [ ] This PR stores or transmits data that was not stored or transmitted before.
- [ ] This PR requires additional review of its security implications for other reasons.

### DIFF
--- a/.github/workflows/build-runner-images.yml
+++ b/.github/workflows/build-runner-images.yml
@@ -33,8 +33,8 @@ jobs:
       - uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
         with:
           params: |
-            AMI_ACCOUNT=/github-actions-runner-ami/account
-            AMI_FILTER=/github-actions-runner-ami/filter
+            AMI_ACCOUNT=/github-runner/gold-image-account
+            AMI_FILTER=/github-runner/gold-image-filter
 
       - name: Retrieve default VPC ID and subnet
         id: vpc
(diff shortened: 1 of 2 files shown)


.github/workflows/build-runner-images.yml
```diff
@@ -33,8 +33,8 @@ jobs:
       - uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
         with:
           params: |
-            AMI_ACCOUNT=/gold-image/account
-            AMI_FILTER=/gold-image/filter
+            AMI_ACCOUNT=/github-actions-runner-ami/account
+            AMI_FILTER=/github-actions-runner-ami/filter
```
gsf, line 37: Looks good but would you mind setting these at `/github-runner/gold-image-account` and `/github-runner/gold-image-filter` to match the path I set up for other github-runner params?

gsf (approved):
As mentioned in Slack, make sure you update /github-runner/ami-account and /github-runner/ami-filter in Parameter Store before merge so that we point at our built image with the apply that drops user data. You could even updated those params and run the plan workflow again to check.

And make sure to test with a self-hosted job as soon as the apply is complete so we know if these images can actually run jobs.

PR #44, opened 2024-01-11
PLT-277: Add Start Script

gfreeman-navapbc:
## 🎫 Ticket

## 🛠 Changes

Adds start script from example to use runners.

## 🔒 Security Implications

None.


gsf:
Could we move the install-runner.sh and start-runner.sh scripts into the packer/github-actions-runner directory?

gsf:
And would you add a README.md to packer/github-actions-runner referencing https://github.com/philips-labs/terraform-aws-github-runner/tree/4f257f9ff643e1f31e597a15dd9a0a65afbdc605/images as the source for scripts, etc.

terraform/services/github-actions-runner/main.tf
```diff
@@ -80,7 +80,7 @@ module "github-actions-runner" {
   role_permissions_boundary = data.aws_iam_policy.developer_boundary_policy.arn
 
   enable_ssm_on_runners = true
-  enable_userdata       = false
+  enable_userdata       = true
```
gsf, line 83: Revert of userdata to true should be in another PR. If anything this PR should switch this back to false.

packer/github-actions-runner/build.pkr.hcl
```diff
@@ -47,6 +47,20 @@ build {
     ]
   }
 
+  provisioner "file" {
+    content = templatefile("./start-runner.sh", {
+      start_runner = templatefile("../start-runner.sh", { metadata_tags = "enabled" })
```
gsf, line 52: Why does the provisioner reference the start-runner.sh script twice here? And why is the second in the parent directory?

gsf: Oh I see it's because the aws-github-runner repo uses the first as a wrapper script. I think we could drop the wrapper and just have a single template file.

gsf (approved):
Looks good. Let's merge, and it would be good if you can test by applying temporary terraform changes from your local so we can avoid the back and forth of PRs to flip that AMI switch until we're sure it works.

PR #51, opened 2024-02-20
PLT-292 Test infra for opt-out-import

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-292

## 🛠 Changes

Refactoring of shared terraform to support test infra for opt-out-import.

## ℹ️ Context for reviewers

In adding the opt-out-import test infra I realized we could wrap resources like S3 buckets and KMS keys in modules to standardize configuration. I also updated handling of AWS account IDs to mask them in logs.

I've also dropped the reusable terraform-command workflow, which is both too powerful (allowing one-off infra changes without understanding context) and not flexible enough to enable updates like the ones mentioned in Acceptance Validation below.

Note that these changes will also require updates to roles and buckets in opt-out-import workflows in ab2d-lambdas and dpc-app, and roles in ab2d and ab2d-events.

## ✅ Acceptance Validation

All terraform plans should pass. I'm also temporarily updating some apply workflows to move resources in state (like tfstate buckets) to avoid deletion. These should be reviewed on merge, and I will follow up with a PR to return those apply workflows to their usual commands.

## 🔒 Security Implications

None.


gfreeman-navapbc (approved):
LGTM

PR #64, opened 2024-03-15
PLT-362 Give AB2D export function max memory

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-362

## 🛠 Changes

Set ab2d opt-out lambda function to max memory (10GB).

## ℹ️ Context for reviewers

AB2D export function currently uses a little under 5GB and runs for 28.6 seconds, but that may grow. Even at 10GB memory, running for 28.6 seconds costs approximately $0.00477, so cost concerns are low. We can adjust downward once usage is more established.

Screenshot 2024-03-15 at 4 42 58 PM

## ✅ Acceptance Validation

See checks.

## 🔒 Security Implications

None.


Sadibhatla (approved):
Thank you so much

PR #70, opened 2024-03-29
PLT-383 Remove sbx config from opt-out

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-383

## 🛠 Changes

Configuration specific to the sbx (sandbox) environment was removed from opt-out terraform.

## ℹ️ Context for reviewers

The opt-out lambdas are not deployed to the sandbox environment, so configuration specific to that environment is not needed.

## ✅ Acceptance Validation

See checks, should be no change in terraform plan.

## 🔒 Security Implications

None.


knavapbc (approved):
lgtm

PR #71, opened 2024-04-04
Increase runner hop limit to 2

gsf:
## 🎫 Ticket

N/A

## 🛠 Changes

Upped hop limit on the runner from 1 to 2.

## ℹ️ Context for reviewers

Preemptively addressing this having seen the issue at https://github.com/philips-labs/terraform-aws-github-runner/issues/3821, which states that `http_put_response_hop_limit` must be increased to 2 for containers to access the instance metadata and access credentials for the instance profile.

## ✅ Acceptance Validation

See checks for image build.

## 🔒 Security Implications

None.

### DIFF
--- a/packer/github-actions-runner/build.pkr.hcl
+++ b/packer/github-actions-runner/build.pkr.hcl
@@ -16,7 +16,7 @@ build {
   provisioner "shell" {
     environment_vars = []
     inline = concat([
-      "sudo yum -y upgrade-minimal",
+      "sudo yum -y update --security",
       "sudo yum -y install amazon-cloudwatch-agent jq git docker",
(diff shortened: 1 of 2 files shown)


gfreeman-navapbc (approved):

gsf:
On the build we're running into an unrelated issue with grub2 installation. Looking into it.

gsf:
I switched from `upgrade-minimal` because that subcommand is only defined for `dnf`, not `yum`, and limited it to security updates. That may have solved the problem, but I still see grub2 in dependencies, so it may just be that Amazon fixed their software repositories for AL2. In any case, it builds now, so I'll merge.

PR #95, opened 2024-06-14
Access-log bucket

christopher-maboh:
## 🎫 Ticket

## 🛠 Changes

Create Access-log bucket

## ℹ️ Context

Our shared terraform currently creates an access log bucket for each bucket. We've recently hit the 100-bucket limit in the BCDA account (quota increase pending) so there is a reason not to create buckets unnecessarily.

## 🧪 Validation

terraform plan ran locally


terraform/services/bucket-access-log/README.md
```diff
@@ -0,0 +1,12 @@
+# Terraform for bucket-access-log in target accounts
```
gsf, line 1: Your naming below convinced me that we should go with the plural "bucket-access-logs" for the name of the service instead. Please rename, sorry!

terraform/services/bucket-access-log/main.tf
```diff
@@ -0,0 +1,53 @@
...
+  bucket = aws_s3_bucket.bucket-access_logs.id
+
+  rule {
+    apply_server_side_encryption_by_default {
+      sse_algorithm = "aws:kms"
+    }
+  }
+}
+
+data "aws_iam_policy_document" "bucket-access_logs" {
```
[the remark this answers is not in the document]
christopher-maboh: Adding this policy to bucket access logs. I have a worry. this part of the policy that's the bucket access logs' name and the path/prefix to the logs
resources = ["arn:aws:s3:::${local.bucket}/${local.s3_logs_prefix}/*"]
my worry is that in the bucket module we are using the bucket name as prefix to logs that is var.name, so am not sure how it will work. I used data source to look up the bucket name in the bucket service but that means the bucket in the module has to be created first for data look up to work and we are also using a data source to look up the access log bucket in the the bucket module

terraform/services/bucket-access-log/README.md
```diff
@@ -0,0 +1,12 @@
+# Terraform for bucket-access-logs in target accounts
```
gsf, line 1: Please modify the directory name for the service to "bucket-access-logs" as well.

terraform/services/bucket-access-log/main.tf
```diff
@@ -0,0 +1,72 @@
...
+      identifiers = ["logging.s3.amazonaws.com"]
+      type        = "Service"
+    }
+
+    actions = [
+      "s3:PutObject"
+    ]
+
+    resources = [
+      "arn:aws:s3:::${aws_s3_bucket.bucket_access_logs.bucket}/${local.s3_logs_prefix}/*"
```
gsf, line 64: Following up on your comment at https://github.com/CMSgov/ab2d-bcda-dpc-platform/pull/95#discussion_r1657915697, we don't need to be that specific in defining resources here. We can set it at the bucket instead of the prefix and drop the s3_logs_path variable.

```suggestion
      "arn:aws:s3:::${aws_s3_bucket.bucket_access_logs.bucket}/*"
```

terraform/services/bucket-access-logs/main.tf
```diff
@@ -0,0 +1,72 @@
+resource "aws_s3_bucket" "bucket_access_logs" {
+  bucket        = "${var.app}-${var.env}-bucket-access-log"
```
gsf, line 2: Since we're creating just one bucket per account, let's use the account ID in the bucket name. Add an [aws_caller_identity data block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/caller_identity) and reference the account ID like so:

```suggestion
  bucket        = "${data.aws_caller_identity.current.account_id}-bucket-access-logs"
```

.github/workflows/bucket-access-log-apply.yml
```diff
@@ -0,0 +1,40 @@
+name: bucket-access-logs apply terraform
```
gsf, line 1: Missed changing the name on this file

terraform/services/bucket-access-logs/terraform.tf
```diff
@@ -0,0 +1,18 @@
+provider "aws" {
+  default_tags {
+    tags = {
+      application = var.app
+      business    = "oeda"
+      code        = "https://github.com/CMSgov/ab2d-bcda-dpc-platform/tree/main/terraform/services/bucket-access-logs"
+      component   = "bucket-access-logs"
+      environment = var.env
+      terraform   = true
```
gsf, line 9: The app and env variables are no longer needed here since the bucket is at the account level. The variables file can then also be removed, and no need to set TF_VAR_app and TF_VAR_env in the workflows.

gsf (approved):
Looks good! Great work

PR #96, opened 2024-06-14
PLT-355 switch resource block to data block

christopher-maboh:
## 🎫 Ticket

## 🛠 Changes

Resource block for access-log removed and data block added to reference access log bucket 

## ℹ️ Context

Our shared terraform currently creates an access log bucket for each bucket. We've recently hit the 100-bucket limit in the BCDA account (quota increase pending) so there is a reason not to create buckets unnecessarily. A terraform service for an access-log-bucket was created and points all bucket-logging to prefixes within that bucket.

## 🧪 Validation

terraform ran locally. only error was Error: reading S3 Bucket (bcda-access-logs): couldn't find resource because infra creating access log bucket is yet to be deployed and is being use here as a data block.

all workflows fail due to the fact that the data source being used is yet to be deployed and so terraform can't find centralized bucket

### DIFF
--- a/terraform/modules/function/main.tf
+++ b/terraform/modules/function/main.tf
@@ -145,7 +145,6 @@ module "zip_bucket" {
     "arn:aws:iam::${data.aws_ssm_parameter.prod_account[0].value}:role/delegatedadmin/developer/${var.app}-prod-github-actions",
     "arn:aws:iam::${data.aws_ssm_parameter.sbx_account[0].value}:role/delegatedadmin/developer/${var.app}-sbx-github-actions",
   ] : []
-  
 }
 
 resource "aws_s3_object" "empty_function_zip" {


terraform/modules/bucket/main.tf
```diff
@@ -86,64 +86,12 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "this" {
...
-  bucket = aws_s3_bucket.access_logs.id
-  policy = data.aws_iam_policy_document.access_logs.json
-}
-
-resource "aws_s3_bucket_versioning" "access_logs" {
-  bucket = aws_s3_bucket.access_logs.id
-  versioning_configuration {
-    status = "Enabled"
-  }
+data "aws_s3_bucket" "bucket-access_logs" {
```
gsf, line 89:
```suggestion
data "aws_s3_bucket" "bucket_access_logs" {
```

terraform/modules/bucket/variables.tf
```diff
@@ -5,6 +5,16 @@ variable "name" {
...
+
+variable "app" {
+  description = "The name of the application"
+  type        = string
+}
+
+variable "env" {
+  description = "The environment name"
+  type        = string
+}
```
gsf, line 20: Please add validation as in https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/6cc555cad64f7cb69d2a417f13d34657316b9c1a/terraform/modules/vpc/variables.tf

terraform/modules/bucket/outputs.tf
```diff
@@ -7,3 +7,13 @@ output "id" {
...
+
+output "access_log_bucket_name" {
+  description = "The name of the access log S3 bucket"
+  value       = data.aws_s3_bucket.bucket-access_logs.bucket
+}
+
+output "access_log_bucket_arn" {
+  description = "The ARN of the access log S3 bucket"
+  value       = data.aws_s3_bucket.bucket-access_logs.arn
+}
```
gsf, line 19: I don't think we need these outputs

terraform/modules/bucket/main.tf
```diff
@@ -86,64 +86,12 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "this" {
...
-  }
-}
-
 resource "aws_s3_bucket_logging" "this" {
   bucket = aws_s3_bucket.this.id
 
-  target_bucket = aws_s3_bucket.access_logs.id
-  target_prefix = "log/"
+  target_bucket = data.aws_s3_bucket.bucket-access_logs.id
+  target_prefix = "${var.name}/log/"
```
gsf, line 96:
```suggestion
  target_prefix = "${var.name}/"
```

terraform/modules/function/main.tf
```diff
@@ -145,6 +145,9 @@ module "zip_bucket" {
     "arn:aws:iam::${data.aws_ssm_parameter.prod_account[0].value}:role/delegatedadmin/developer/${var.app}-prod-github-actions",
     "arn:aws:iam::${data.aws_ssm_parameter.sbx_account[0].value}:role/delegatedadmin/developer/${var.app}-sbx-github-actions",
   ] : []
+  
+  app = var.app
+  env = var.env
```
gsf, line 150: These will also need to be added here: https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/6cc555cad64f7cb69d2a417f13d34657316b9c1a/terraform/services/tfstate/main.tf#L7

The tfstate workflows should also be updated to include bucket and table module paths.

gsf:
Looks good! We'll have to get #95 merged and applied before we get good plan checks here.

terraform/modules/bucket/main.tf
```diff
@@ -86,64 +86,12 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "this" {
...
-  policy = data.aws_iam_policy_document.access_logs.json
-}
-
-resource "aws_s3_bucket_versioning" "access_logs" {
-  bucket = aws_s3_bucket.access_logs.id
-  versioning_configuration {
-    status = "Enabled"
-  }
+data "aws_s3_bucket" "bucket_access_logs" {
+  bucket = "${var.app}-${var.env}-bucket-access-log"
```
gsf, line 90: With the change suggested in https://github.com/CMSgov/ab2d-bcda-dpc-platform/pull/95/files#r1674341305 we'll be able to drop app and env variables and reference the account ID as we did there.

terraform/modules/function/main.tf
```diff
@@ -145,6 +145,7 @@ module "zip_bucket" {
     "arn:aws:iam::${data.aws_ssm_parameter.prod_account[0].value}:role/delegatedadmin/developer/${var.app}-prod-github-actions",
     "arn:aws:iam::${data.aws_ssm_parameter.sbx_account[0].value}:role/delegatedadmin/developer/${var.app}-sbx-github-actions",
   ] : []
+
```
gsf, line 148: Sorry to nit pick but could this blank line addition be removed?

christopher-maboh: blank line addition has been removed

gsf (approved):
Looks good! Let's monitor the apply jobs

PR #99, opened 2024-06-21
BCDA-8208: Rename BCDA lambda handlers to bootstrap

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8208

## 🛠 Changes

- Rename handlers to bootstrap

## ℹ️ Context

Was looking at the lambda logs and saw some errors with trying to find the entrypoint. After reading the AWS documentation, I realized that it expects the zip file to have exactly one binary, which is specifically named bootstrap 😅

This was an oversight on my part — I thought if the handler was specified as cclf-import / opt-out-import here, it would look for that binary filename instead.

## 🧪 Validation

Changed manually in AWS Console and confirmed successful lambda initialization.

### DIFF
--- a/terraform/services/cclf-import/main.tf
+++ b/terraform/services/cclf-import/main.tf
@@ -33,7 +33,7 @@ module "cclf_import_function" {
   name        = local.full_name
   description = "Ingests the most recent CCLF from BFD"
 
-  handler = "cclf-import"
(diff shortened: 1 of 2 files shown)


gsf (approved):
Don't know if this came up with DPC but I thought there was a reason we stuck with bootstrap.

PR #101, opened 2024-06-26
PLT-348: Shared WAF Terraform Module

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-348

## 🛠 Changes

Added the Terraform code for the shared WAF module.

## ℹ️ Context

We're trying to consolidate the varying ways our APIs handle whitelisting to streamline IP address management.

## 🧪 Validation

Verified that terraform plan runs and succeeds for all APIs and environments


gsf:
Agreed on the need for a module for WAF (which I would consider calling "firewall"), but I would approach this by adding terraform services in this PR as well. I think each team has a firewall on their main API as well as the website.

terraform/modules/firewall/main.tf
```diff
@@ -0,0 +1,172 @@
...
+
+resource "aws_wafv2_web_acl_logging_configuration" "this" {
+  log_destination_configs = [aws_cloudwatch_log_group.this.arn]
+  resource_arn = aws_wafv2_web_acl.this.arn
+
+  depends_on = [
+    aws_wafv2_web_acl.this,
+    aws_cloudwatch_log_group.this
+  ]
+}
```
gsf, line 162: Let's hold on logging configuration for now. Looks like current WAFs send to an account S3 bucket via Firehose.

terraform/services/api-waf/main.tf
```diff
@@ -0,0 +1,33 @@
...
+    }
+    dpc = {
+      name = "${var.app}-${var.env}-1"
+    }
+  }
+}
+
+data "aws_lb" "api_lb" {
+  name = local.load_balancers[var.app].name
+}
```
gsf, line 22: This is the right idea, getting load balancers in the service via data calls.

terraform/modules/firewall/main.tf
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  visibility_config {
+    cloudwatch_metrics_enabled = true
+    metric_name                = "WAF_Common_Protections"
+    sampled_requests_enabled   = true
+  }
+
+  dynamic "rule" {
+    for_each = var.rate_based_rule != null ? [var.rate_based_rule] : []
```
gsf, line 30: Let's try setting a rate-based rule here that we can generally use across WAFs instead of making it all dynamic. Base it on the BCDA rate limiting.

terraform/modules/firewall/main.tf
```diff
@@ -0,0 +1,172 @@
...
+      visibility_config {
+        cloudwatch_metrics_enabled = true
+        metric_name                = rule.value.name
+        sampled_requests_enabled   = true
+      }
+    }
+  }
+
+  dynamic "rule" {
+    for_each = var.ip_sets_rule
```
gsf, line 70: I think we can make assumptions about IP sets here. If the WAF has an IP sets rule, it should be set to an empty IP set since we'll be adding ranges outside of this terraform.

gsf (commented):
Generally looks good, just nudging in the direction of fewer variables and less dynamic configuration. We can always extend as needs are realized.

gfreeman-navapbc (commented):
LGTM, can't leave an approval since I started the PR

gsf (approved):
Okay let's merge

PR #102, opened 2024-07-08
add slack notification on terraform apply fail

knavapbc:
## 🎫 Ticket


gsf (commented):
Yes this generally looks like the right direction. I think a plain text "slack-message" stating that a terraform apply has failed and including a link to the failing job should suffice. We'll want to keep these short since it's possible we'll receive a number of them at once if an apply fails across multiple envs.

PR #105, opened 2024-07-22
PLT-460 Add sg-ip-rules script

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-460

## 🛠 Changes

Added sg-ip-rules script.

## ℹ️ Context

I've added this sg-ip-rules script as an example and something quick modifications could be made to in order to set security group rules quickly from the command line.

## 🧪 Validation

Used in filling in security group rules for zscaler-private and zscaler-public groups for PLT-460.

### DIFF
--- a/scripts/sg-ip-rules
+++ b/scripts/sg-ip-rules
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+if { [ -z "$1" ]; }; then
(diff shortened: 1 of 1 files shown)


ildesenesence (approved):
LGTM - I really like the elegant use of `shift` and the `for` loop to capture the CIDRs in this script

PR #106, opened 2024-07-23
set batch_size to 1

christopher-maboh:
## 🎫 Ticket

## 🛠 Changes
batch_size set to 10

## ℹ️ Context

Terraform failed upon apply with error related to batch_size

### DIFF
--- a/terraform/modules/queue/main.tf
+++ b/terraform/modules/queue/main.tf
@@ -70,6 +70,6 @@ resource "aws_sns_topic_subscription" "this" {
 resource "aws_lambda_event_source_mapping" "this" {
   event_source_arn = aws_sqs_queue.this.arn
   function_name    = var.function_name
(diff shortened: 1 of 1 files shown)


gsf:
Looks like the only change to infra is here: https://github.com/CMSgov/ab2d-bcda-dpc-platform/actions/runs/10064444177/job/27821368181?pr=106#step:6:85

Would you mind checking with the BCDA team and asking if the batch size needs to remain at 1?

gsf (approved):
Looks good! As I mentioned in Slack, we can parameterize this if we identify a need for higher size batch sizes for future functions and queues.

christopher-maboh:
ok Sean

> Looks good! As I mentioned in Slack, we can parameterize this if we identify a need for higher size batch sizes for future functions and queues.

ok Sean

PR #107, opened 2024-07-26
PLT-338 Update Github Actions Runners to support node20

christopher-maboh:
## 🎫 Ticket

[PLT-338](https://jira.cms.gov/browse/PLT-338)

## 🛠 Changes

- https://github.com/CMSgov/ab2d-bcda-dpc-platform/pull/104 was reverted
- ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION was set to true in the workflow
- packer build was updated to use al2023-legacy-gi-2024-07-10T10-40-52Z as the base image

## ℹ️ Context

The most [recent automated build](https://github.com/CMSgov/ab2d-bcda-dpc-platform/actions/runs/7865662073) of the Github Actions Runner AMI failed due to an update in the [hashicorp/setup-packer](https://github.com/hashicorp/setup-packer) action

## 🧪 Validation

See checks for packer build.

### DIFF
--- a/packer/github-actions-runner/sources.pkr.hcl
+++ b/packer/github-actions-runner/sources.pkr.hcl
@@ -9,7 +9,9 @@ source "amazon-ebs" "github-actions-runner" {
   iam_instance_profile                      = "bcda-mgmt-github-actions"
 
   source_ami_filter {
-     filters = { name = "${var.ami_filter}" }
+     filters = {
+      name = "al2023-legacy-gi-*"
+     }
     owners = ["${var.ami_account}"]
     most_recent = true
   }
--- a/packer/github-actions-runner/sources.pkr.hcl
+++ b/packer/github-actions-runner/sources.pkr.hcl
@@ -9,7 +9,9 @@ source "amazon-ebs" "github-actions-runner" {
   iam_instance_profile                      = "bcda-mgmt-github-actions"
 
   source_ami_filter {
-    filters = { name = "${var.ami_filter}" }
+     filters = {
+      name = "al2023-legacy-gi-2024-07-10T10-40-52Z" 
+    }
     owners = ["${var.ami_account}"]
     most_recent = true
   }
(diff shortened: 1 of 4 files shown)


packer/github-actions-runner/sources.pkr.hcl
```diff
@@ -9,7 +9,9 @@ source "amazon-ebs" "github-actions-runner" {
   iam_instance_profile                      = "bcda-mgmt-github-actions"
 
   source_ami_filter {
-    filters = { name = "${var.ami_filter}" }
+     filters = {
+      name = "al2023-legacy-gi-2024-07-10T10-40-52Z"
```
gsf, line 13: The filter should allow for new images to be selected each month:

```suggestion
      name = "al2023-legacy-gi-*" 
```

Also, since we're hardcoding the filter here instead of referencing `var.ami_filter` (which I'm in favor of doing) we can drop the `PKR_VAR_ami_filter` param in the workflow and remove the variable from packer/github-actions-runner/variables.pkr.hcl.

gsf (commented):
I think we'll want to set `remote_folder` to "/home/ec2-user" to address the "Permission denied" error at https://github.com/CMSgov/ab2d-bcda-dpc-platform/actions/runs/10116992800/job/27981049308#step:9:31

I found this suggestion: https://github.com/hashicorp/packer/issues/2875

And it appears Amazon recommends /home/ec2-user: https://aws.amazon.com/blogs/containers/building-amazon-linux-2-cis-benchmark-amis-for-amazon-eks/

packer/github-actions-runner/sources.pkr.hcl
```diff
@@ -9,7 +9,7 @@ source "amazon-ebs" "github-actions-runner" {
   iam_instance_profile                      = "bcda-mgmt-github-actions"
 
   source_ami_filter {
-    filters = { name = "${var.ami_filter}" }
+     filters = { name = "${var.ami_filter}" }
```
gsf, line 12: The check in the PR is still building with the AL2 image instead of AL2023 because /github-runner/gold-image-filter in Parameter Store is set to "amzn2legacy-gi-*". As noted in my earlier comments, I'd rather hardcode the filter string (now "al2023-legacy-gi-*") here and avoid passing it in as a variable.

gsf (approved):
Looks good! Thanks for working through the issues with /tmp scripts. As mentioned in standup, let's try another PR following this without the `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION` env var to check whether the AL2023 image does indeed let us run node20.

PR #109, opened 2024-07-31
PLT-338 Remove actions to allow use of unsecured node version

christopher-maboh:
## 🎫 Ticket

[PLT-338](https://jira.cms.gov/browse/PLT-338)

## 🛠 Changes

ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION  environment variable was removed

## 🧪 Validation
removing ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION env var to check whether the AL2023 image does indeed let us run node20.

### DIFF
--- a/.github/workflows/build-runner-images.yml
+++ b/.github/workflows/build-runner-images.yml
@@ -20,8 +20,6 @@ jobs:
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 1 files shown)


gsf:
This change was included in #110.

PR #111, opened 2024-08-07
BCDA-8301: Increase CCLF import memory size to 3072MB

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8301

## 🛠 Changes

- Ensure CCLF import has enough memory to process files

## ℹ️ Context

This is a short-term resolution for issues we are seeing with our CCLF import lambda being killed for certain large files. These files are _not_ close to meeting the 1024MB size on their own, but they are downloaded four times within the process.

We will prioritize a fix to ensure that these files are only downloaded once during processing.

https://jira.cms.gov/browse/BCDA-8302

## 🧪 Validation

Increased manually and verified that the lambda could process all the files we were having issues with.
CleanShot 2024-08-07 at 16 52 46@2x

### DIFF
--- a/terraform/services/cclf-import/main.tf
+++ b/terraform/services/cclf-import/main.tf
@@ -9,7 +9,7 @@ locals {
     bcda = "bcda-${var.env}-rds"
   }
   memory_size = {
-    bcda = 1024
+    bcda = 2048
   }
 }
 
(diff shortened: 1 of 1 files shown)


kyeah:
I actually just merged https://github.com/CMSgov/bcda-app/pull/985 which reduces the memory consumption, so I'll go ahead and close this one out + revert the memory in AWS Console 😄

PR #116, opened 2024-08-28
PLT-585 External services IP sets

gsf:
## 🎫 Ticket

## 🛠 Changes

Add WAF IP sets for external services, including Zscaler and New Relic.

## ℹ️ Context

In setting up the new WAF for DPC dev, we realized the need for shared IP sets within an account that would contain IPs for external services. This terraform creates those IP sets but leaves management of the addresses/ranges within them up to other processes (manual for now, automated from CIDRs in the private repo in the future).

## 🧪 Validation

See plans in checks.

### DIFF
--- a/terraform/services/external-services-ip-sets/main.tf
+++ b/terraform/services/external-services-ip-sets/main.tf
@@ -1,0 +1,32 @@
+# The address used is from the 203.0.113.0/24 range reserved for
+# documentation, which should be unresolvable in any network. It
+# is only used as a placeholder, and should be overwritten by
+# other processes when managing addresses in these IP sets.
+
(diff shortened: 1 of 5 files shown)


terraform/services/external-services-ip-sets/main.tf
```diff
@@ -0,0 +1,32 @@
+# The address used is from the 203.0.113.0/24 range reserved for
+# documentation, which should be unresolvable in any network. It
+# is only used as a placeholder, and should be overwritten by
+# other processes when managing addresses in these IP sets.
```
gfreeman-navapbc, line 4: LGTM, clever workaround here

gfreeman-navapbc (approved):

PR #118, opened 2024-08-29
PLT-585 IP sets for the api-waf service

gsf:
## 🎫 Ticket

## 🛠 Changes

Added the ability to set IP sets in the firewall module, and added an IP set for customers in the api-waf service. Also referenced the external-services IP set created in #116 and #117.

## ℹ️ Context

This will associate terraform-managed IP sets with the WAF for each API (starting with DPC dev environment). CIDRs in these IP sets will be managed outside of terraform.

## 🧪 Validation

See plans in checks.

### DIFF
--- a/.github/workflows/api-waf-apply.yml
+++ b/.github/workflows/api-waf-apply.yml
@@ -1,7 +1,9 @@
 name: api-waf plan terraform
 
 on:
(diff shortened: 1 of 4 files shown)


knavapbc (approved):
lgtm

PR #120, opened 2024-09-03
PLT-338 Security group for internet access

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-338

## 🛠 Changes

Switched zscaler-security-groups to service-security-groups in terraform services and added a security group for internet access. Also extended plan and apply of these security groups to include ab2d and dpc mgmt environments.

## ℹ️ Context

In working on Packer issues I realized we had no standard security group for internet egress. The service-security-groups terraform service should also be useful for adding other general-purpose security groups. 

## 🧪 Validation

See plans in checks.


ildesenesence (approved):
No functional or style recommendations - looks good to me

PR #121, opened 2024-09-05
PLT-585: Add override to aws-common ruleset for XSS check

gfreeman-navapbc:
## 🎫 Ticket

## 🛠 Changes

This adds an override to the AWSManagedRulesCommonRuleSet's CrossSiteScripting_BODY check to allow DPC to send HTML in the requests to certain endpoints of their API.

## ℹ️ Context

This should fix the failing smoke tests DPC has in the dev environment. We found that this rule had been blocking requests to certain endpoints.

## 🧪 Validation

Terraform plan should pass and the DPC smoke test in DEV should pass once this is applied.

### DIFF
--- a/.github/workflows/api-waf-apply.yml
+++ b/.github/workflows/api-waf-apply.yml
@@ -7,6 +7,7 @@ on:
     paths:
       - .github/workflows/api-waf-apply.yml
       - terraform/services/api-waf/**
+      - terraform/modules/firewall/**
   workflow_dispatch: # Allow manual trigger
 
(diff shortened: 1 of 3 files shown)


gsf (approved):
Looks great

gsf:
Sorry, change looks good, but we should see a plan check in the PR. Please add `terraform/modules/firewall/**` to paths for api-waf-plan and api-waf-apply workflows.

gsf (approved):

PR #122, opened 2024-09-05
PLT-585: Override size requirement WAF rule

gfreeman-navapbc:
## 🎫 Ticket

## 🛠 Changes

Adds an override block for the `SizeRestrictions_BODY` AWS provided WAF rule.

## ℹ️ Context

After moving to batch requests to avoid hitting the rate limit WAF rule, we foudn that the requests would be too large for AWS' provided WAF rules.

## 🧪 Validation

DPC smoke tests should pass after this is applied.

### DIFF
--- a/terraform/modules/firewall/main.tf
+++ b/terraform/modules/firewall/main.tf
@@ -115,6 +115,14 @@ resource "aws_wafv2_web_acl" "this" {
             count {}
           }
         }
+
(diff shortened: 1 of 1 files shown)


gsf (approved):
Plan at https://github.com/CMSgov/ab2d-bcda-dpc-platform/actions/runs/10728339143/job/29752686178#step:6:26 looks good!

PR #124, opened 2024-09-10
PLT-532: Gitleaks Precommit for MBI Scanning

ildesenesence:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-532

## 🛠 Changes

Added .gitleaks.toml config file for pre-commit testing and detection of MBIs as defined by https://www.cms.gov/medicare/new-medicare-card/understanding-the-mbi.pdf

## ℹ️ Context

PLT-532 was opened in order to facilitate a guardrail against developers committing changes with PHI/PII in the form of MBI numbers following an incident in 2023. Gitleaks pre-commit hook is a method of doing so programmatically, with a minimum of developer manual processing.

## 🧪 Validation

Changes were validated against a randomly generated file with strings in MBI format, and a .json formatted report was generated to demonstrate the types of findings that would be presented by a gitleaks precommit hook.


gsf:
Looks great so far. In the interest of not cluttering up the root of the repo, could any of the files go in a subdirectory?

.gitignore
```diff
@@ -7,3 +7,4 @@ node_modules
 terraform.tfstate*
 terraform.tfvars
 .terraform
+.gitleaks.toml
```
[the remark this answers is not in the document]
ildesenesence: Yes, I had mistakenly named it with a leading dot which automatically added it to .gitignore. This has been removed in the current state of the branch

ildesenesence:
> @gsf : Looks great so far. In the interest of not cluttering up the root of the repo, could any of the files go in a subdirectory?

Moved the gitleaks.toml to ./ops and referenced the new location in the .pre-commit-config.yaml; I appreciate the suggestion because while it does still create an additional entry in the folder, if we have future elements we want to include, or a future revision where the pre-commit automatically generates a report .json file, we have a location to put it.

gsf:
I'm going to contradict myself a bit here but I think I'd prefer we keep the .gitleaks.toml in its standard location at the root of the repo, especially since we'll be propagating this out to all repos and may need to update existing .gitleaks.toml configurations in some.

mjburling (commented):
This is a fine first step from which to iterate. Very curious to see where this ends up.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,6 @@
 repos:
   - repo: https://github.com/gitleaks/gitleaks
-    rev: v8.16.1
+    rev: v8.18.4
     hooks:
       - id: gitleaks
+        args: [--config ./ops/gitleaks.toml]
```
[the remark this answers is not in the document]
ildesenesence: Appreciate this feedback, probably going to be irrelevant though as I'm agreeing with @gsf that for supportability's sake, I'd rather keep with the pattern of having gitleaks.toml in the root of the repo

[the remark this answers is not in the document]
mjburling: Even better.

gsf:
The gitleaks.toml filename still needs a dot in front of it, correct?

ildesenesence:
No, that's only if we want it to be ignored by github, which we don't. Default config is `gitleaks.toml`

gitleaks.toml
```diff
@@ -0,0 +1,8 @@
+title = "MBI Scan"
+# Only using custom rules
+useDefault = false
```
gsf, line 3: I think we want to keep the defaults enabled to continue scanning for other secrets on commit.

ildesenesence:
@gsf I would normally agree, however that would expand the scope of this ticket to also creating a baseline for existing secrets in the git log that we're willing to accept as either no longer present in the current code, or acceptable due to them being sample data, test data, or inactive in the system. If we want to expand the scope to include default scanning, I'm not explicitly _against_ that, but it does expand the scope of both this proof of concept and the subsequent implementation significantly.

gsf:
Because we have not specified a config before now, we already use the default secret scanning in our pre-commit hook. If we don't extend the default then we would be dropping that.

ildesenesence:
> Because we have not specified a config before now, we already use the default secret scanning in our pre-commit hook. If we don't extend the default then we would be dropping that.

You're absolutely right. Thank you for the catch - I missed my own notes on why I chose this repo to begin with, as it already had the initial gitleaks setup/baseline checking done. Should be updated to restore that now.

mjburling:
> No, that's only if we want it to be ignored by github, which we don't. Default config is `gitleaks.toml`

As of gitleaks v8.18.4, without overrides as environment variables or explicit (`-s`/`--source`) or (`-c`/`--config`) flags, gitleaks will use `./.gitleaks.toml` or it will use the [_default configuration_](https://github.com/gitleaks/gitleaks/blob/master/config/gitleaks.toml). Config location (`-c`/`--config`) is divined thusly:

>  1. --config/-c
>  2. env var GITLEAKS_CONFIG
>  3. (--source/-s)/.gitleaks.toml
>
>    If none of the three options are used, then gitleaks will use the default config

As of v8.19.x. (since 2024-09-13), it looks like they've made some... less modest adjustments, deprecating `detect` and `protect` and it seems that (`-s`/`--source`) was caught up in that. The third bullet now reads:

> 3. (target path)/.gitleaks.toml

Seems as though "target path" is the positional argument to `gitleaks dir` and `gitleaks git` subcommands.

ildesenesence:
> > No, that's only if we want it to be ignored by github, which we don't. Default config is `gitleaks.toml`
> 
> As of gitleaks v8.18.4, without overrides as environment variables or explicit (`-s`/`--source`) or (`-c`/`--config`) flags, gitleaks will use `./.gitleaks.toml` or it will use the [_default configuration_](https://github.com/gitleaks/gitleaks/blob/master/config/gitleaks.toml). Config location (`-c`/`--config`) is divined thusly:
> 
> > 1. --config/-c
> > 2. env var GITLEAKS_CONFIG
> > 3. (--source/-s)/.gitleaks.toml
> > 
> > If none of the three options are used, then gitleaks will use the default config
> 
> As of v8.19.x. (since 2024-09-13), it looks like they've made some... less modest adjustments, deprecating `detect` and `protect` and it seems that (`-s`/`--source`) was caught up in that. The third bullet now reads:
> 
> > 3. (target path)/.gitleaks.toml
> 
> Seems as though "target path" is the positional argument to `gitleaks dir` and `gitleaks git` subcommands.

truly an "are you kidding me" moment. I'll review and see how it impacts this work.

gitleaks.toml
```diff
@@ -0,0 +1,8 @@
+title = "MBI Scan"
+# Only using custom rules
```
gsf, line 2: Thanks for switching to extend the default config! This comment can be dropped, and I think `useDefault` needs to be in the `[extend]` section as shown at https://github.com/gitleaks/gitleaks?tab=readme-ov-file#configuration

gitleaks.toml
ildesenesence: Reverted this in most recent version, closing

gitleaks.toml
ildesenesence: Pushing correction now

.gitleaks.toml
```diff
@@ -0,0 +1,10 @@
+title = "MBI Scan"
+# Only using custom rules
+
```
mjburling, line 3: Unsure if this will be enduring, but _something_ general seems more appropriate here.
```suggestion
title = "DASG Standard"

```

.gitleaks.toml
```diff
@@ -0,0 +1,10 @@
...
+[extend]
+useDefault = true
```
mjburling, line 5: Maybe we'll have a pre-commit function in place for toml one day. Ahead of that, I've just been using the `taplo` cli+lsp for sanity checking. Based on the `[[rules]]` formatting below, it looks like we _might_ prefer something like `indent_entries=true` in e.g. `taplo fmt --option indent_entries=true .gitleaks.toml`

```suggestion
  useDefault = true
```

.gitleaks.toml
ildesenesence: Given that this is a proof of concept and the naming of the config file is an easy lift, I think I'm going to prefer reserving a choice like this until we're creating a central config to deploy across multiple repositories

gsf: I think it is worth setting a more general title here. Let's keep it very boring and call it "gitleaks config".

gsf (approved):

PR #123, opened 2024-09-10
importing ab2d-env rds

knavapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-562=3

## 🛠 Changes

no change expected

## ℹ️ Context

Changes on the .tf files to make sure after import, terrafor plan do not show any breaking changes

## 🧪 Validation
comment will show the output of terraform plan

PR #125, opened 2024-09-11
PLT-563 Apply api-rds terraform to AB2D

knavapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-563

## 🛠 Changes

Updates to api-rds terraform service following changes in #123.

## ℹ️ Context

Changes on the .tf files to make sure after import, terrafor plan do not show any breaking changes

## 🧪 Validation

See checks for terraform plan.

### DIFF
--- a/.github/workflows/api-rds-apply.yml
+++ b/.github/workflows/api-rds-apply.yml
@@ -1,22 +1,16 @@
-name: api-rds plan terraform
+name: api-rds apply terraform
 
 on:
-  pull_request:
+  push:
+    branches:
+      - main
     paths:
-      - .github/workflows/api-rds-plan.yml
+      - .github/workflows/api-rds-apply.yml
       - terraform/services/api-rds/**
   workflow_dispatch: # Allow manual trigger
 
 jobs:
-  check-terraform-fmt:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: ./actions/setup-tfenv-terraform
-      - run: terraform fmt -check -diff -recursive terraform/services/api-rds
-
-  terraform-plan:
-    needs: check-terraform-fmt
+  terraform-apply:
     permissions:
       contents: read
       id-token: write
@@ -27,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        app: [ab2d, bcda, dpc]
+        app: [ab2d]
         env: [dev, test, sbx, prod]
     steps:
       - uses: actions/checkout@v4
@@ -37,7 +31,7 @@ jobs:
           role-to-assume: arn:aws:iam::${{ matrix.app == 'ab2d' && secrets[format('{0}_{1}_ACCOUNT', matrix.app, matrix.env)] || secrets.BCDA_ACCOUNT }}:role/delegatedadmin/developer/${{ matrix.app }}-${{ matrix.env }}-github-actions
           aws-region: ${{ vars.AWS_REGION }}
       - run: terraform init -backend-config=../../backends/${{ matrix.app }}-${{ matrix.env }}.s3.tfbackend
-      - run: terraform plan
+      - run: terraform apply -auto-approve
         env:
           TF_VAR_app: ${{ matrix.app }}
           TF_VAR_env: ${{ matrix.env }}
--- a/.github/workflows/api-rds-apply.yml
+++ b/.github/workflows/api-rds-apply.yml
@@ -34,4 +34,3 @@ jobs:
       - run: terraform apply -auto-approve
         env:
           TF_VAR_app: ${{ matrix.app }}
-          TF_VAR_env: ${{ matrix.env }}
--- a/.github/workflows/api-rds-apply.yml
+++ b/.github/workflows/api-rds-apply.yml
@@ -31,6 +31,12 @@ jobs:
           role-to-assume: arn:aws:iam::${{ matrix.app == 'ab2d' && secrets[format('{0}_{1}_ACCOUNT', matrix.app, matrix.env)] || secrets.BCDA_ACCOUNT }}:role/delegatedadmin/developer/${{ matrix.app }}-${{ matrix.env }}-github-actions
           aws-region: ${{ vars.AWS_REGION }}
       - run: terraform init -backend-config=../../backends/${{ matrix.app }}-${{ matrix.env }}.s3.tfbackend
+      - uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
+        env:
+          AWS_REGION: ${{ vars.AWS_REGION }}
+        with:
+          params: |
+            TF_VAR_jenkins_security_group_id=/jenkins/security-group
       - run: terraform apply -auto-approve
         env:
           TF_VAR_app: ${{ matrix.app }}
--- a/terraform/backends/ab2d-prod-test.s3.tfbackend
+++ b/terraform/backends/ab2d-prod-test.s3.tfbackend
@@ -1,0 +1,2 @@
+bucket         = "ab2d-prod-test-tfstate"
+dynamodb_table = "ab2d-prod-test-tfstate"
--- a/terraform/backends/ab2d-prod-test.s3.tfbackend
+++ b/terraform/backends/ab2d-prod-test.s3.tfbackend
@@ -1,2 +1,0 @@
-bucket         = "ab2d-prod-test-tfstate"
-dynamodb_table = "ab2d-prod-test-tfstate"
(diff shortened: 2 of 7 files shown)


terraform/services/api-rds/main.tf
```diff
@@ -41,8 +41,5 @@ resource "aws_db_instance" "api" {
     Name             = "${local.db_name[var.app][var.env]}-rds"
     "cpm backup"     = "Monthly"
     contact          = "ab2d-ops@semanticbits.com"
```
gsf, line 43: Remove these tags as well

gsf:
https://github.com/CMSgov/ab2d-bcda-dpc-platform/pull/123#discussion_r1755315263

terraform/services/api-rds/variables.tf
```diff
@@ -8,10 +8,10 @@ variable "app" {
 }
 
 variable "env" {
-  description = "The application environment (dev, test, sbx, prod)"
+  description = "The application environment (dev, test, sbx, prod-test, prod)"
   type        = string
   validation {
-    condition     = contains(["dev", "test", "sbx", "prod"], var.env)
+    condition     = contains(["dev", "test", "sbx", "prod-test", "prod"], var.env)
     error_message = "Valid value for env is dev, test, sbx, or prod."
```
gsf, line 15: Include prod-test in the error message.

gsf:
Add an api-rds-apply.yml workflow file that only includes AB2D environments.

gsf:
We can hold off on adding the prod-test environment for now. Remove all prod-test references, and instead focus on adding resources and the apply workflows in this PR.

knavapbc:
@gsf final review 🙏

terraform/backends/ab2d-prod-test.s3.tfbackend
```diff
@@ -0,0 +1,2 @@
+bucket         = "ab2d-prod-test-tfstate"
+dynamodb_table = "ab2d-prod-test-tfstate"
```
gsf, line 2: Drop this backend file. No prod-test env for now.

terraform/services/api-rds/main.tf
```diff
@@ -1,10 +1,11 @@
...
     ab2d = {
-      dev  = "ab2d-dev"
-      test = "ab2d-east-prod-test"
-      sbx  = "ab2d-sbx-sandbox"
-      prod = "ab2d-east-prod"
+      dev       = "ab2d-dev"
+      test      = "ab2d-east-impl"
+      sbx       = "ab2d-sbx-sandbox"
+      prod      = "ab2d-east-prod"
+      prod-test = "ab2d-east-prod-test"
```
gsf, line 8: Drop prod-test

terraform/services/api-rds/variables.tf
```diff
@@ -8,10 +8,10 @@ variable "app" {
...
 
 variable "env" {
-  description = "The application environment (dev, test, sbx, prod)"
+  description = "The application environment (dev, test, sbx, prod-test, prod)"
   type        = string
   validation {
-    condition     = contains(["dev", "test", "sbx", "prod"], var.env)
-    error_message = "Valid value for env is dev, test, sbx, or prod."
+    condition     = contains(["dev", "test", "sbx", "prod-test", "prod"], var.env)
+    error_message = "Valid value for env is dev, test, sbx, prod-test, or prod."
```
gsf, line 15: Drop prod-test

gsf (changes requested):
The terraform still needs to be extended to cover resources in the ab2d data terraform as requested in this comment: https://github.com/CMSgov/ab2d-bcda-dpc-platform/pull/123#discussion_r1755315263

knavapbc:
@gsf

.github/workflows/api-rds-apply.yml
```diff
@@ -0,0 +1,37 @@
...
+      - uses: ./actions/setup-tfenv-terraform
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::${{ matrix.app == 'ab2d' && secrets[format('{0}_{1}_ACCOUNT', matrix.app, matrix.env)] || secrets.BCDA_ACCOUNT }}:role/delegatedadmin/developer/${{ matrix.app }}-${{ matrix.env }}-github-actions
+          aws-region: ${{ vars.AWS_REGION }}
+      - run: terraform init -backend-config=../../backends/${{ matrix.app }}-${{ matrix.env }}.s3.tfbackend
+      - run: terraform apply -auto-approve
+        env:
+          TF_VAR_app: ${{ matrix.app }}
+          TF_VAR_env: ${{ matrix.env }}
```
gsf, line 37: Mind setting your editor to end files with a newline?

gsf (commented):
Exactly the right direction with the added resources. As we just discussed, I'd like to keep required variables at the service level limited to "app" and "env" if at all possible. I think I'd still prefer to keep resources in the service if you can set defaults for the other variables or define them in locals instead. Otherwise, I could see the benefit in creating a more parameterized "database" module that's called from the service.

ildesenesence:
> I'd like to keep required variables at the service level limited to "app" and "env" if at all possible. I think I'd still prefer to keep resources in the service if you can set defaults for the other variables or define them in locals instead. Otherwise, I could see the benefit in creating a more parameterized "database" module that's called from the service.

I think I'm going to lean more service-only if possible - the backend configuration files from the ab2d-ops repository have additional fields compared to the backends\*.tfstate files in the platform repo and I'm still trying to figure out what potholes those might introduce.

terraform/services/api-rds/terraform.tf
```diff
@@ -7,6 +7,7 @@ provider "aws" {
       component   = "api-rds"
       environment = var.env
       terraform   = true
+      contact     = "ab2d-ops@semanticbits.com"
```
gsf, line 10: We can drop the "contact" tag

terraform/services/api-rds/variables.tf
```diff
@@ -1,9 +1,15 @@
...
 variable "app" {
   description = "The application name (ab2d, bcda, dpc)"
   type        = string
   validation {
+    condition     = contains(["ab2d"], var.app)
+    error_message = "Valid value for app is ab2d."
+    /*
     condition     = contains(["ab2d", "bcda", "dpc"], var.app)
     error_message = "Valid value for app is ab2d, bcda, or dpc."
+*/
```
gsf, line 12: Let's leave bcda and dpc in the validation. They can still be in the validation but not included in the apply workflow, and this will be one less thing to change once we're applying in those environments.

terraform/services/api-rds/main.tf
```diff
@@ -2,47 +2,130 @@ locals {
...
-      test = "${var.app}-${var.env}"
-      sbx  = "${var.app}-${var.env}"
-      prod = "${var.app}-${var.env}"
-    }
-    dpc = {
-      dev  = "${var.app}-${var.env}"
-      test = "${var.app}-${var.env}"
-      sbx  = "${var.app}-${var.env}"
-      prod = "${var.app}-${var.env}"
-    }
```
gsf, line 20: Could we leave something in here for bcda and dpc so the plan checks pass?

terraform/services/api-rds/terraform.tf
ildesenesence: Done

terraform/services/api-rds/variables.tf
ildesenesence: Readded to the validation in latest push

terraform/services/api-rds/main.tf
ildesenesence: Added a variation on this in the latest push - will look at updating the locals to set this up for success in the other environments on testing

gfreeman-navapbc (commented):
Can we remove bcda and dpc here? IMO if they're out of scope we should just add them when we need to, rather than have failing plans in the meantime

ildesenesence:
> Can we remove bcda and dpc here? IMO if they're out of scope we should just add them when we need to, rather than have failing plans in the meantime

Honestly that does feel like a reasonable and straightforward solution here.

.github/workflows/api-rds-apply.yml
```diff
@@ -0,0 +1,42 @@
...
+      id-token: write
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./terraform/services/api-rds
+    strategy:
+      fail-fast: false
+      matrix:
+        app: [ab2d]
+        env: [dev, test, sbx, prod]
```
gsf, line 25: Let's remove all but the test environment for now to only apply there with this merge. That should resolve PLT-691, and we can add the others back in in a future PR after validating in that environment.

ildesenesence:
Closed as redundant - Development on this branch can continue for PLT-563, but PLT-691 obviates some of these changes and this will need to be rebased and extended for the other ab2d environments, along with a PR for ab2d-ops to cease managing those environments following the format used in https://github.com/CMSgov/ab2d-bcda-dpc-platform/pull/134

PR #127, opened 2024-09-13
DPC-3767 Add infrastructure for DPC WAF-sync lambda

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3767

## 🛠 Changes

- Added wav-sync as service
- Added plan and apply wav-sync github actions

## ℹ️ Context

DPC needs to synchronize ip addresses in a database with an AWS WAF IP Set. We [decided](https://confluence.cms.gov/x/1ZUZQQ) to use a cron-triggered lambda to manage this. This ticket creates the infrastructure for automated deployment of the lambda.

The aws_waf_access inline policy seems to be the minimum necessary for the lambda to run. It may require updating (the lambda has not been written).

The name of the IP Set was intuited by browsing the WAF IP Sets in AWS Console.

The terraform scripts and github actions are modified versions of the opt-out-import service.

## 🧪 Validation

- Comparison of the plan output to opt-out-import plan output show that the permissions necessary are present.
- Will apply the plan manually after the first round of code reviews.


gfreeman-navapbc:
@gsf I wonder if it's a good idea to rename these files to be DPC-specific, since I am unsure if the other teams will be following a similar route to mange the WAF IP-sets. Although, it could be a good idea to push BCDA and AB2D to use this automated method as well

terraform/services/waf-sync/main.tf
```diff
@@ -0,0 +1,66 @@
...
+  from_port   = 5432
+  to_port     = 5432
+  protocol    = "tcp"
+  description = "waf-sync function access"
+
+  security_group_id        = data.aws_security_group.db.id
+  source_security_group_id = module.waf_sync_function.security_group_id
+}
+
+# Because we inline policies, we cannot just link to aws:policy/AWSWAFFullAccess
```
gsf, line 54: We can drop the waf_role data block up above, right?

gsf (commented):
Nicely done overall. We can leave dpc-specific references for now but I would like to call it "api-waf-sync" to avoid changing the name if bcda moves to the same architecture.

gsf (approved):
Looks good, and the plan looks good as well

jdettmannnava:
> Looks good, and the plan looks good as well

Except AWS has it's own custom cron syntax...

PR #128, opened 2024-09-17
PLT-503: Add test and sandbox environments to DPC WAF plan and apply

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-503

## 🛠 Changes

Adds test and sbx environments to the DPC WAF configuration

## ℹ️ Context

These are changes made as a part of the overall WAF migration work.

## 🧪 Validation

Once this is applied, we should see the Web ACL configurations show up with the placeholder IP sets. We'll need to manually update them in AWS, and then reapply once we remove the association with the security group on ingress.

### DIFF
--- a/.github/workflows/api-waf-plan.yml
+++ b/.github/workflows/api-waf-plan.yml
@@ -4,7 +4,6 @@ on:
   pull_request:
     paths:
       - .github/workflows/api-waf-plan.yml
-      - .github/workflows/api-waf-apply.yml
       - terraform/services/api-waf/**
       - terraform/modules/firewall/**
   workflow_dispatch: # Allow manual trigger
--- a/.github/workflows/api-waf-plan.yml
+++ b/.github/workflows/api-waf-plan.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - .github/workflows/api-waf-plan.yml
+      - .github/workflows/api-waf-apply.yml
       - terraform/services/api-waf/**
       - terraform/modules/firewall/**
   workflow_dispatch: # Allow manual trigger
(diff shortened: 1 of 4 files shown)


ildesenesence (approved):
Curious what impact our engineers should actually see with this - seems like this is just setting up the skeleton for rules to be capable of being applied, but isn't actually applying rules? I haven't followed this all the way back to the current WAF configuration to verify that.

gsf (changes requested):
The api-waf service will need to be updated to not create and associate IP sets on sandbox.

gfreeman-navapbc:
> Curious what impact our engineers should actually see with this - seems like this is just setting up the skeleton for rules to be capable of being applied, but isn't actually applying rules? I haven't followed this all the way back to the current WAF configuration to verify that.

See https://github.com/CMSgov/ab2d-bcda-dpc-platform/tree/main/terraform/modules/firewall
and https://github.com/CMSgov/ab2d-bcda-dpc-platform/tree/main/terraform/services/api-waf

.github/workflows/api-waf-plan.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - .github/workflows/api-waf-plan.yml
+      - .github/workflows/api-waf-apply.yml
```
gsf, line 7: The apply workflow doesn't need to be included here

gfreeman-navapbc: Yea, this was only to kick off the initial plan runs because it wasn't picking up on the first commit that only included the apply changes, should have removed once I started changing other files

gsf (approved):

PR #129, opened 2024-09-27
PLT-504: Add BCDA dev environment to WAF configuration

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-504

## 🛠 Changes

This adds the BCDA dev environment to the apply workflow of the new WAF configuration.

## ℹ️ Context

We're moving all apps to WAF for IP filtering. This should alleviate some of the issues we have with security groups and maintaining our CIDR whitelists.

## 🧪 Validation

BCDA dev should be able to be reached on the VPN, but no other traffic should be let through. The BCDA team may need to add CIDR ranges to the new IP-set that will be created. We'll also need to remove the association to the dev WAF and security group in place in the bcda-ops repo: https://github.com/CMSgov/bcda-ops/pull/1114

### DIFF
--- a/.github/workflows/api-waf-apply.yml
+++ b/.github/workflows/api-waf-apply.yml
@@ -24,6 +24,9 @@ jobs:
       matrix:
         app: [dpc]
         env: [dev, test, sbx]
+        include:
+          - app: bcda
+            env: dev
     steps:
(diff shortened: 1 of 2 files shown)


gsf (approved):
Looks good!

PR #130, opened 2024-09-27
PLT-663 Connect to packer instances via session manager

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-663

## 🛠 Changes

Updates Packer sources to connect via session manager, and cleans up variables in packer.

## ℹ️ Context

This PR originally updated workflow and packer to allow for building images on github-hosted runners, but stepped back from that approach while we work out compliance around usage of GitHub-hosted runners. I've kept the changes to connect to the instance via session manager as this improves on the process regardless of where the packer is being run.

By running the build for runner images on github-hosted images, we avoid the cyclical dependency that keeps us from building a new image if the runner image is broken. Also, builds are faster because we skip the startup time of self-hosted runners.

## 🧪 Validation

See checks for successful build.


ildesenesence (approved):
lgtm - the github diff was a little disorienting in sources.bkr.hcl but once I realized it was matching up lines weird because github was incorrectly lining up an add as if it were a modify, it became much more legible. Great work

gsf:
> lgtm - the github diff was a little disorienting in sources.bkr.hcl but once I realized it was matching up lines weird because github was incorrectly lining up an add as if it were a modify, it became much more legible. Great work

Yes more changes there than would be otherwise due to HCL formatting.

gsf:
Closing this PR. I've merged changes I'll keep into #110.

PR #132, opened 2024-10-09
PLT-650: Add BCDA to apply matrix for WAF config

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-650

## 🛠 Changes

Adds BCDA to the apply workflow and updates a comment with a link to the decision log

## ℹ️ Context

We're continuing our rollout of the new WAF config to the BCDA test and sandbox environments.

## 🧪 Validation

Smoke tests should succeed in each environment and network traffic should remain the same. To be merged after https://github.com/CMSgov/bcda-ops/pull/1118 is approved and deployed to the environments.

### DIFF
--- a/.github/workflows/api-waf-apply.yml
+++ b/.github/workflows/api-waf-apply.yml
@@ -22,11 +22,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        app: [dpc]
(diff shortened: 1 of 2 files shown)


gsf (approved):
Matrix looks good, and thanks for linking that doc

PR #133, opened 2024-10-10
PLT-634: Add DPC PROD env to WAF apply

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-634

## 🛠 Changes

Adds DPC Prod environment to the new WAF config apply

## ℹ️ Context

We rolled out these changes to DEV, TEST, and SBX and through troubleshooting in those environments, believe these changes to be ready for the DPC Prod rollout.

## 🧪 Validation

Will verify traffic from accepted IP ranges is still allowed, and blocked otherwise. Needs to be merged and applied after https://github.com/CMSgov/dpc-ops/pull/741


.github/workflows/api-waf-apply.yml
```diff
@@ -27,6 +27,8 @@ jobs:
         include:
           - app: bcda
             env: dev
+          - app: dpc
+            env: prod
```
gsf, line 31: I think this could be added to the env list on line 26?

gfreeman-navapbc: Yea, wasn't sure whether BCDA test and sbx environments would be deployed first, so kept it in the include, but it seems like this will go first, so I will make that change

MEspositoE14s (approved):
👍

gfreeman-navapbc:
> I think this could be added to the env list on line 26?

RE: The environment matrix: Sean had me take dpc prod out of the include and put it into the matrix as we hadn't gotten around to deploying the BCDA WAF changes yet. However, since we've now deployed those changes to dev, test, and sbx, we need to separate the production deploy necessary for DPC while keeping the BCDA config in place.

ildesenesence (approved):
LGTM

PR #134, opened 2024-10-11
PLT-691 Apply api rds service to ab2d test

ildesenesence:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-691

## 🛠 Changes

Added management of AB2D Test environment to api-rds service in Platform repository
Imported existing resources to state file to track managed components

## ℹ️ Context

Changes were made to centralize database management under PLT-691 and PLT-563

## 🧪 Validation

Clean terraform plans will resolve locally and will be shown in GHA tasks


terraform/services/api-rds/main.tf
```diff
@@ -2,47 +2,132 @@ locals {
...
+    apply_method = "immediate"
+  }
+  parameter {
+    name         = "shared_preload_libraries"
+    value        = "pg_stat_statements,pg_cron"
+    apply_method = "pending-reboot"
+  }
+  parameter {
+    name         = "cron.database_name"
+    value        = local.db_name
```
gsf, line 82: I just realized that this does need to be a short name for ab2d databases, which is why they have [this map](https://github.com/CMSgov/ab2d-ops/blob/396c802e95381555230ee64043ecdeac597d968b/terraform/modules/data/main.tf#L2C3-L7) in ab2d-ops. Luckily, the short name is pretty close to our envs, so I think we can manage it like so:

```suggestion
    value        = var.app == "ab2d" && var.env == "test" ? "impl" : var.env
```

ildesenesence: That makes sense! We can talk about why this is when I get back but for now I think this handles it fine to be functional. :)

gsf (approved):
Looks good!

PR #135, opened 2024-10-11
PLT-691 Apply api-rds service to ab2d-test

ildesenesence:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-691

## 🛠 Changes

Patch to add TF_VAR_env to apply workflow

## ℹ️ Context

Continuing PLT-691 implementation

## 🧪 Validation

Compared to existing workflows, and arose from an error in applying PR #134


gsf (approved):
That should do it 🤞🏼

PR #140, opened 2024-10-30
PLT-724 Apply api-rds service to ab2d dev

ildesenesence:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-724

## 🛠 Changes

Added dev environment to plan/apply action matrices

## ℹ️ Context

PLT-712 needs for us to update ab2d dev and test to Postgres 16, this is attempting to make that change from the same location. for both .

## 🧪 Validation

Validated locally with terraform plan; also verifying via terraform plan github action


gfreeman-navapbc (approved):
lgtm

PR #142, opened 2024-11-01
PLT-724 fixed parameter error for rds.logical_replication

ildesenesence:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-724 

## 🛠 Changes

Changing rds.logical_replication from immediate to pending-reboot apply method

## ℹ️ Context

AWS/Hashicorp documentation was unclear that the parameter was static and could not be applied immediately, and verification via plan/github action was misleading as the plan showed no errors, but the apply did error.

## 🧪 Validation

Successful plan locally, validated all "immediate" apply method parameters referenced are dynamic by postgres documentation

### DIFF
--- a/terraform/services/api-rds/main.tf
+++ b/terraform/services/api-rds/main.tf
@@ -109,7 +109,7 @@ resource "aws_db_parameter_group" "parameter_group" {
   }
   parameter {
     name         = "rds.logical_replication"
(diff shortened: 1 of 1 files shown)


oluwolenpbc (approved):

ildesenesence:
Reverting to draft - Need answers on whether we also need logical replication disabled in this PR before it can be implemented.

PR #143, opened 2024-11-01
Override NoUserAgent_HEADER block on WAF config

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-635

## 🛠 Changes

Adds an override for requests lacking the User-Agent HTTP header

## ℹ️ Context

BCDA receives a majority of their requests from automated systems that lack this header, and do not require it in the previous iteration of the WAF.

## 🧪 Validation

Requests with no user-agent header (eg., scheduled FHIR scan requests) should pass through the WAF.

### DIFF
--- a/terraform/modules/firewall/main.tf
+++ b/terraform/modules/firewall/main.tf
@@ -112,14 +112,6 @@ resource "aws_wafv2_web_acl" "this" {
         name        = "AWSManagedRulesCommonRuleSet"
         vendor_name = "AWS"
 
-        # Override for XSS block on request body, DPC team sends HTML blocks in requests to certain endpoints
-        rule_action_override {
-          name = "CrossSiteScripting_BODY"
-          action_to_use {
-            count {}
-          }
-        }
-
         # Override for size requirements of requests, this is set at 8kb which is too small for some acceptable requests
         rule_action_override {
           name = "SizeRestrictions_BODY"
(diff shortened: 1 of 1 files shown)


terraform/modules/firewall/main.tf
```diff
@@ -112,19 +112,33 @@ resource "aws_wafv2_web_acl" "this" {
...
+          for_each = var.app == "dpc" ? ["apply"] : []
+          content {
+            name = "CrossSiteScripting_BODY"
+            action_to_use {
+              count {}
+            }
+          }
+        }
+
+        # Override for requests lacking a User-Agent header, as most BCDA requests are automated and lack them
```
ildesenesence, line 134: Is there a good way for us to detect when a request is automated vs. not? This is probably the realm of enhancements but wanted to ask the question.

ildesenesence (approved):
Looks good overall

terraform/modules/firewall/main.tf
gfreeman-navapbc: Not really, the only relatively "good" sign I can see through cursory research is the lack of this header, but BCDA doesn't currently require it, so rather than break all the existing calls to the API we'd rather override for the moment and have the BCDA team see if it's something that they'd like to require moving forward

PR #144, opened 2024-11-08
PLT-674 redux: adding Athena queries

timothy-ellis-ky:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-674

## 🛠 Changes

Expansion of PLT-674 POC for DPC to include Athena queries that should support Quicksight Dashboard

## ℹ️ Context

Proof of Concept for DPC Quicksight Dashboard. This PR attempts to establish the following in DEV and TEST:
1. create Firehose subscription that would transform and export DPC Cloudwatch data to an S3 bucket for streams /aws/ecs/fargate/dpc-{ENV}-aggregation and /aws/ecs/fargate/dpc-{ENV}-api
2. create Glue tables to receive transformed Cloudwatch data
3. enable GH Action workflows that will plan and apply changes accordingly.
4. add S3 Result bucket and Athena queries against the Glue tables
5. provide acceptable access permissions across all resources

## 🧪 Validation

All resources except latest additions (S3 + Athena) are available in DEV; adjustment of the GH Action matrix settings presume to support all environments; recommend incremental expansion of matrix.env to support necessary adjustments.


gsf:
Great work, and thanks for testing the QuickSight plan out in this PoC. Is any of this live in our infrastructure currently?

One note is that the Context section says "DEV only" but the workflows include dev and test.

Also I would prefer to call the service "quicksight" instead of "dpc-quicksights" to allow for extending this for the other APIs.

timothy-ellis-ky:
> Great work, and thanks for testing the QuickSight plan out in this PoC. Is any of this live in our infrastructure currently?
> 
> One note is that the Context section says "DEV only" but the workflows include dev and test.
> 
> Also I would prefer to call the service "quicksight" instead of "dpc-quicksights" to allow for extending this for the other APIs.

1. PLT-674 as demonstrated Nov 7 is live in DEV; removal and resubmission of PR adds support for TEST but GHA will not deploy until merge.
2. Open to suggestions on naming conventions and modularity; PR is specific to DPC, unclear where to generalize (other projects not yet analyzed) at this time.

gsf:
No need to generalize for ab2d and bcda at this point but I would like to leave the opportunity open to extend to those APIs without requiring a rename of the service. It's a bit of a pain to update references across files in this PR but now is the time to do so. 

I'm fine with this being dev and test. The PR description should be updated to clarify.

timothy-ellis-ky:
> No need to generalize for ab2d and bcda at this point but I would like to leave the opportunity open to extend to those APIs without requiring a rename of the service. It's a bit of a pain to update references across files in this PR but now is the time to do so.
> 
> I'm fine with this being dev and test. The PR description should be updated to clarify.

PR Description updated

terraform/services/quicksight/s3.tf
```diff
@@ -0,0 +1,102 @@
...
+
+resource "aws_s3_bucket_server_side_encryption_configuration" "dpc-insights-athena" {
+  bucket = aws_s3_bucket.dpc-insights-athena.id
+  rule {
+    apply_server_side_encryption_by_default {
+      kms_master_key_id = local.this_env_key
+      sse_algorithm     = "aws:kms"
+    }
+  }
+}
```
gsf, line 102: I think this file could be replaced by a reference to the bucket module. See this reference for example: https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/main/terraform/services/tfstate/main.tf#L5-L8

terraform/services/quicksight/firehose-agg.tf
```diff
@@ -0,0 +1,109 @@
+# Firehose Data Stream
+resource "aws_kinesis_firehose_delivery_stream" "firehose-ingester-agg" {
```
gsf, line 2: The resource name doesn't need to repeat the resource, and we should stick to underscores:

```suggestion
resource "aws_kinesis_firehose_delivery_stream" "ingester_agg" {
```

terraform/services/quicksight/lambda.tf
```diff
@@ -0,0 +1,38 @@
+data "archive_file" "zip-archive-format-dpc-logs" {
+  type        = "zip"
+  source_file = "${path.module}/lambda_src/dpc-bfd-cwlog-basic-flatten-json.py"
+  output_path = "${path.module}/lambda_src/${local.this_env}/dpc-bfd-cwlog-basic-flatten-json.zip"
+}
+
+# Lambda Function to process logs from Firehose
+resource "aws_lambda_function" "lambda-function-format-dpc-logs" {
```
gsf, line 8:
```suggestion
resource "aws_lambda_function" "format_dpc_logs" {
```

terraform/services/quicksight/firehose-api.tf
```diff
@@ -0,0 +1,102 @@
+# Firehose Data Stream
+resource "aws_kinesis_firehose_delivery_stream" "firehose-ingester-api" {
```
gsf, line 2:
```suggestion
resource "aws_kinesis_firehose_delivery_stream" "ingester_api" {
```

terraform/services/quicksight/glue.tf
```diff
@@ -0,0 +1,274 @@
...
+  lifecycle {
+    ignore_changes = [
+      # TODO: Consider removing everything here
+      parameters
+    ]
+  }
+}
+
+# add crawler for metadata inspection
+resource "aws_glue_crawler" "glue_crawler_api_metrics" {
```
gsf, line 234:
```suggestion
resource "aws_glue_crawler" "api_metrics" {
```

terraform/services/quicksight/terraform.tf
```diff
@@ -0,0 +1,18 @@
...
+      component   = "dpc-quicksights"
+      environment = var.env
+      terraform   = true
+    }
+  }
+}
+
+terraform {
+  backend "s3" {
+    key = "dpc-quicksights/terraform.tfstate"
```
gsf, line 16:
```suggestion
    key = "quicksight/terraform.tfstate"
```

terraform/services/quicksight/terraform.tf
```diff
@@ -0,0 +1,18 @@
+provider "aws" {
+  default_tags {
+    tags = {
+      application = var.app
+      business    = "oeda"
+      code        = "https://github.com/CMSgov/ab2d-bcda-dpc-platform/tree/main/terraform/services/dpc-quicksights"
+      component   = "dpc-quicksights"
```
gsf, line 7:
```suggestion
      code        = "https://github.com/CMSgov/ab2d-bcda-dpc-platform/tree/main/terraform/services/quicksight"
      component   = "quicksight"
```

gsf (commented):
Looking good, and I don't want to hold it up too much longer, but see the naming suggestions and the potential for using the existing bucket module.

gsf (approved):
Looks good! Thanks for updating the naming and incorporating the bucket module.

PR #146, opened 2024-11-13
PLT-728 update state and tf to postgres upgrade

ildesenesence:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-728

## 🛠 Changes

Updating TF to match current AWS state per Postgres 16 upgrade work plan

## ℹ️ Context

Due to technical issues with Postgres 16.4 upgrade there were some settings changed via console and this pull is intended to realign terraform with the intended settings.

## 🧪 Validation

Local TF plan & TF refresh-only updates for state file; also validating via GHA plan

### DIFF
--- a/terraform/services/api-rds/main.tf
+++ b/terraform/services/api-rds/main.tf
@@ -160,17 +160,17 @@ resource "aws_db_parameter_group" "v16_parameter_group" {
   }
   parameter {
     name         = "max_logical_replication_workers"
-    values       = "22"
+    value        = "22"
     apply_method = "pending-reboot"
   }
   parameter {
     name         = "max_worker_processes"
-    values       = "25"
+    value        = "25"
     apply_method = "pending-reboot"
   }
   parameter {
     name         = "max_wal_senders"
-    values       = "20"
+    value        = "20"
     apply_method = "pending-reboot"
   }
 
--- a/terraform/services/api-rds/main.tf
+++ b/terraform/services/api-rds/main.tf
@@ -92,21 +92,6 @@ resource "aws_db_parameter_group" "v16_parameter_group" {
     value        = 0
     apply_method = "pending-reboot"
   }
-  parameter {
-    name         = "max_logical_replication_workers"
-    value        = "22"
-    apply_method = "pending-reboot"
-  }
-  parameter {
-    name         = "max_worker_processes"
-    value        = "25"
-    apply_method = "pending-reboot"
-  }
-  parameter {
-    name         = "max_wal_senders"
-    value        = "20"
-    apply_method = "pending-reboot"
-  }
 
   lifecycle {
     create_before_destroy = true
--- a/terraform/services/api-rds/main.tf
+++ b/terraform/services/api-rds/main.tf
@@ -11,8 +11,8 @@ locals {
   }[var.app]
   postgres_ver = {
     ab2d = {
-      dev  = 15
-      test = 15
+      dev  = 16
+      test = 16
       sbx  = 15
       prod = 15
     }[var.env]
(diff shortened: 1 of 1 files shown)


terraform/services/api-rds/main.tf
```diff
@@ -158,7 +103,7 @@ resource "aws_db_parameter_group" "v16_parameter_group" {
 resource "aws_db_instance" "api" {
   allocated_storage   = 500
   engine              = "postgres"
-  engine_version      = local.postgres_ver
+  engine_version      = 16
```
gsf, line 106: Should we set this to 16.4? I see that's a change in the plan, and I think pinning to minor version is a good idea anyway.

terraform/services/api-rds/main.tf
```diff
@@ -144,7 +89,7 @@ resource "aws_db_parameter_group" "v16_parameter_group" {
   }
   parameter {
     name         = "rds.logical_replication"
-    value        = 0 # contains(["ab2d-dev", "ab2d-east-impl"], local.db_name) ? "1" : "0" # To support blue-green deployment for PostGres16 upgrade
+    value        = 0
     apply_method = "pending-reboot"
   }
```
gsf, line 95: Why were the parameters that will be deleted removed? Are they necessary?

terraform/services/api-rds/main.tf
ildesenesence: It actually doesn't pin to the minor version - we have the default auto_minor_version_upgrade setting (enabled) which is why the dbprotect scan findings of us being 1 revision back resolved themselves automatically on 11/5 (our first scheduled AWS maintenance window after 15.7/14.13 were released for auto-upgrade)

Setting 16 vs 16.4 isn't a meaningful gesture - the actual current version is stored in engine_version_current, which will remain on 16.4

terraform/services/api-rds/main.tf
ildesenesence: If you're talking about the v15 group, I removed that per our conversation; if you're talking about the text past the # in line 147 (92 after changes) then basically I was leaving that as a shortcut if we needed it for blue/green deployments but since we're not doing those it's unnecessary clutter

gsf: I mean in the v16 group, resulting in the parameters removed in the plan:

```
  # aws_db_parameter_group.v16_parameter_group will be updated in-place
  ~ resource "aws_db_parameter_group" "v16_parameter_group" {
        id           = "ab2d-dev-rds-parameter-group-v16"
        name         = "ab2d-dev-rds-parameter-group-v16"
        tags         = {}
        # (5 unchanged attributes hidden)

      - parameter {
          - apply_method = "pending-reboot" -> null
          - name         = "max_logical_replication_workers" -> null
          - value        = "22" -> null
        }
      - parameter {
          - apply_method = "pending-reboot" -> null
          - name         = "max_wal_senders" -> null
          - value        = "20" -> null
        }
      - parameter {
          - apply_method = "pending-reboot" -> null
          - name         = "max_worker_processes" -> null
          - value        = "25" -> null
        }
```

terraform/services/api-rds/main.tf
gsf: So this line in the plan will not result in a database reboot in the next maintenance window?

```
  # aws_db_instance.api will be updated in-place
  ~ resource "aws_db_instance" "api" {
      ~ engine_version                        = "16.4" -> "16"
```

terraform/services/api-rds/main.tf
ildesenesence: Got it - This was a convo between me and @oluwolenpbc where he stated that with logical replication turned off as we're not doing blue-green deployments, these parameters are not needed moving forward.

terraform/services/api-rds/main.tf
ildesenesence: Correct

terraform/services/api-rds/main.tf
gsf: Thanks for explaining! I think this looks good then

gsf (approved):
Nicely done in tracking updates

PR #148, opened 2024-11-14
PLT-674 Second Rd Apply Fixes - Athena

timothy-ellis-ky:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-674

## 🛠 Changes

change Athena database bucket reference from ARN to ID

## ℹ️ Context

Previous branch PLT-674-fix failed to apply completely

## 🧪 Validation

see above


timothy-ellis-ky:
manual run of "quicksight apply" workflow succeeded; these resources are now available in DEV and TEST

gsf (approved):
Looks good! One way to get these before merge would be to run apply on the test environment for this branch with the manual workflow dispatch.

PR #149, opened 2024-11-15
PLT-563 Updated workflows to add sbx and prod

ildesenesence:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-563

## 🛠 Changes

Added sbx and prod to workflow matrices for apply and plan

## ℹ️ Context

Following postgres 16 upgrade, this is intended to move management of RDS in ab2d fully into the platform repo

## 🧪 Validation

Local terraform plan & will verify with GHA terraform plan

### DIFF
--- a/.github/workflows/api-rds-apply.yml
+++ b/.github/workflows/api-rds-apply.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         app: [ab2d]
-        env: [dev, test]
+        env: [dev, test, sbx, prod]
     steps:
       - uses: actions/checkout@v4
       - uses: ./actions/setup-tfenv-terraform
--- a/.github/workflows/api-rds-plan.yml
+++ b/.github/workflows/api-rds-plan.yml
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         app: [ab2d]
-        env: [dev, test]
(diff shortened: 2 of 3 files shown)


gsf (approved):
Change looks good but I wanted to check on one small thing in the sandbox plan output:

```
      ~ vpc_security_group_ids                = [
          - "sg-0cb11c4f23a67e825",
```

Looks like this security group (ab2d-sbx-sandbox-db-internal-access-sg) allows access from an internal range. I can't think of a reason we would need access to the sbx db outside that provided in the ab2d-sbx-sandbox-database-sg group. Would you remove this security group after apply to clean things up?

ildesenesence:
> Change looks good but I wanted to check on one small thing in the sandbox plan output:
> 
> ```
>       ~ vpc_security_group_ids                = [
>           - "sg-0cb11c4f23a67e825",
> ```
> 
> Looks like this security group (ab2d-sbx-sandbox-db-internal-access-sg) allows access from an internal range. I can't think of a reason we would need access to the sbx db outside that provided in the ab2d-sbx-sandbox-database-sg group. Would you remove this security group after apply to clean things up?

Looking into this now. I'll compare to current state and see if I need an environment-specific delta in the service.

gsf (approved):
Thanks for looking into that security group! You can remove it in the work you're doing now or it can wait and get cleaned up with PLT-700.

PR #150, opened 2024-11-19
PLT-634: API WAF - Add BCDA PROD to environment matrix

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-634

## 🛠 Changes

Adds BCDA PROD to environment matrix for the shared WAF module rollout.

## ℹ️ Context

This will enable the WAF in the PROD environment and filter all incoming traffic.

## 🧪 Validation

After https://github.com/CMSgov/bcda-ops/pull/1128 is merged, we should be able to merge this PR to enable the WAF module in production. I'll then check for traffic/connection issues, and fail forward (create new PRs to address) if necessary.

### DIFF
--- a/.github/workflows/api-waf-apply.yml
+++ b/.github/workflows/api-waf-apply.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     paths:
-      - .github/workflows/api-waf-*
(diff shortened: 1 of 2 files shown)


ildesenesence (approved):
LGTM

PR #151, opened 2024-11-20
PLT-662: Establish Athena DataSets 

timothy-ellis-ky:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-662

## 🛠 Changes

- Update commentary and definition for Athena Saved Queries (create or replace view)
- Update IAM permissions to include QuickSight actions where appropriate, including S3 CMK access

## ℹ️ Context

These changes were made to enable manual creation of QuickSight data sets, Analyses, and Dashboards

## 🧪 Validation

See Screen shots in Jira Ticket

### DIFF
--- a/terraform/services/quicksight/main.tf
+++ b/terraform/services/quicksight/main.tf
@@ -31,7 +31,7 @@ locals {
   dpc_athena_bucket_key_alias = module.dpc_insights_athena.key_alias
   dpc_athena_bucket_key_arn   = module.dpc_insights_athena.key_arn
   dpc_athena_bucket_id        = module.dpc_insights_data.id
-  
+
   dpc_athena_results_folder_key = "workgroups/${aws_athena_workgroup.quicksight.name}/"
 }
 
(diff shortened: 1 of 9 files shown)


terraform/services/quicksight/main.tf
```diff
@@ -45,4 +48,10 @@ module "dpc_insights_data" {
 module "dpc_insights_athena" {
   source = "../../modules/bucket"
   name   = local.dpc_athena_s3_name
+}
+
+resource "aws_s3_object" "folder" {
+  bucket       = module.dpc_insights_athena.id
+  content_type = "application/x-directory"
+  key          = local.dpc_athena_results_folder_key
```
gsf, line 56: I'm a little unclear as to why this resource is needed. Is there an error if the output_location is supplied without this?

timothy-ellis-ky: TF definitely argued when I didn't have a specific Result_Configuration folder of this object type for the Athena Workgroup resource; I pulled over the comparable config from BFD-Insights

gsf: Still seems odd to me. I don't see any docs about the need to define an x-directory object in S3 for Athena. But it's fine to leave it in for now.

gsf (approved):
Great work

PR #153, opened 2024-12-06
PLT-505: Add AB2D DEV and TEST to WAF-apply environment matrix

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-505

## 🛠 Changes

Adds AB2D target environments to the matrix for deployment.

## ℹ️ Context

We have to add the environments to the matrix in a way that ensures we keep the existing environments' for other teams in place.

## 🧪 Validation

We will deploy this after the changes in the AB2D ops repo are deployed: https://github.com/CMSgov/ab2d-ops/pull/300

### DIFF
--- a/.github/workflows/api-waf-apply.yml
+++ b/.github/workflows/api-waf-apply.yml
@@ -27,6 +27,10 @@ jobs:
         include:
           - app: dpc
             env: prod
+          - app: ab2d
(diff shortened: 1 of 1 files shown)


gsf (approved):
Good way to go about it

PR #154, opened 2024-12-11
Bump DPC rate limit

gfreeman-navapbc:
## 🎫 Ticket

https://cmsgov.slack.com/archives/C04UG13JF9B/p1733939290774959

## 🛠 Changes

Bumping DPC rate limit to 3000requests/5min

## ℹ️ Context

DPC is seeing some customers get timed out

## 🧪 Validation

DPC rate limit rule should be changed, all others should stay the same

### DIFF
--- a/terraform/services/api-waf/main.tf
+++ b/terraform/services/api-waf/main.tf
@@ -52,6 +52,7 @@ module "aws_waf" {
   content_type = "APPLICATION_JSON"
 
   associated_resource_arn = data.aws_lb.api.arn
+  rate_limit = var.app == "dpc" ? 3000 : 300
   ip_sets = var.env == "sbx" ? [] : [
     one(data.aws_wafv2_ip_set.external_services).arn,
     one(aws_wafv2_ip_set.api_customers).arn,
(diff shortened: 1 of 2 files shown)


terraform/services/api-waf/main.tf
```diff
@@ -52,6 +52,7 @@ module "aws_waf" {
   content_type = "APPLICATION_JSON"
 
   associated_resource_arn = data.aws_lb.api.arn
+  rate_limit              = var.app == "dpc" ? 3000 : 300
```
gsf, line 55: This is good but seeing as BCDA is the team with the need for strict rate limits it might be better to flip it. So update this as follows and set the default at https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/a57fd68067485a38805d4e405a94d730fd82abe3/terraform/modules/firewall/variables.tf#L51 to 3000 as well. 

```suggestion
  rate_limit              = var.app == "bcda" ? 300 : 3000
```

gsf (approved):
Looks good!

PR #157, opened 2024-12-30
BCDA-8360: Update WAF sync service for BCDA

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8360

## 🛠 Changes

Update WAF Sync service to allow for BCDA

## ℹ️ Context

Update WAF Sync service to allow for BCDA

### DIFF
--- a/terraform/services/api-waf-sync/main.tf
+++ b/terraform/services/api-waf-sync/main.tf
@@ -1,6 +1,7 @@
 locals {
   full_name  = "${var.app}-${var.env}-api-waf-sync"
-  db_sg_name = "${var.app}-${var.env}-db"
+  bcda_sg_name = "bcda-${var.env}-rds"
+  dpc_sg_name = "dpc-${var.env}-db"
 }
 
 module "api_waf_sync_function" {
@@ -32,7 +33,7 @@ module "api_waf_sync_function" {
 # Add a rule to the database security group to allow access from the function
 
 data "aws_security_group" "db" {
-  name = local.db_sg_name
+  name = "local.${var.app}_sg_name"
 }
 
 resource "aws_security_group_rule" "function_access" {
--- a/terraform/services/api-waf-sync/main.tf
+++ b/terraform/services/api-waf-sync/main.tf
@@ -25,9 +25,9 @@ module "api_waf_sync_function" {
   schedule_expression = "cron(0/10 * * * ? *)"
 
   environment_variables = {
-    ENV          = var.env
-    APP_NAME     = "${var.app}-${var.env}-api-waf-sync"
-    DB_HOST      = "data.aws_ssm_parameter.dpc_db_host.value"
+    ENV      = var.env
+    APP_NAME = "${var.app}-${var.env}-api-waf-sync"
+    DB_HOST  = "data.aws_ssm_parameter.dpc_db_host.value"
   }
 }
 
--- a/terraform/services/api-waf-sync/main.tf
+++ b/terraform/services/api-waf-sync/main.tf
@@ -22,10 +22,10 @@ module "api_waf_sync_function" {
   schedule_expression = "cron(0/10 * * * ? *)"
 
   environment_variables = {
-    ENV             = var.env
-    APP_NAME        = "${var.app}-${var.env}-api-waf-sync"
-    DB_HOST         = "data.aws_ssm_parameter.dpc_db_host.value"
-    DB_URL          = "data.aws_ssm_parameter.bcda_db_url.value"
+    ENV      = var.env
+    APP_NAME = "${var.app}-${var.env}-api-waf-sync"
+    DB_HOST  = "data.aws_ssm_parameter.dpc_db_host.value"
+    DB_URL   = "data.aws_ssm_parameter.bcda_db_url.value"
   }
 }
 
--- a/terraform/services/api-waf-sync/main.tf
+++ b/terraform/services/api-waf-sync/main.tf
@@ -28,7 +28,7 @@ module "api_waf_sync_function" {
     ENV          = var.env
     APP_NAME     = "${var.app}-${var.env}-api-waf-sync"
     DB_HOST      = "data.aws_ssm_parameter.dpc_db_host.value"
-    DATABASE_URL = "data.aws_ssm_parameter.bcda_db_url.value"
+    DATABASE_URL = var.bcda_database_url
   }
 }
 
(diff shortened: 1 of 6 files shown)


terraform/services/api-waf-sync/main.tf
```diff
@@ -22,10 +22,10 @@ module "api_waf_sync_function" {
   schedule_expression = "cron(0/10 * * * ? *)"
 
   environment_variables = {
-    ENV             = var.env
-    APP_NAME        = "dpc-${var.env}-api-waf-sync"
-    WAF_IP_SET_NAME = "DPC_${upper(var.env)}_Implementer_IP_Set"
```
gsf, line 27: Will WAF_IP_SET_NAME no longer be needed? If it is, a map of env to name in the locals could be a good solution.

terraform/services/api-waf-sync/main.tf
```diff
@@ -22,10 +22,10 @@ module "api_waf_sync_function" {
...
-    DB_HOST         = data.aws_ssm_parameter.dpc_db_host.value
+    ENV      = var.env
+    APP_NAME = "${var.app}-${var.env}-api-waf-sync"
+    DB_HOST  = "data.aws_ssm_parameter.dpc_db_host.value"
+    DB_URL   = "data.aws_ssm_parameter.bcda_db_url.value"
```
gsf, line 28: Possible to pick one of DB_HOST and DB_URL and just make one data call?

gsf (commented):
Nicely done

terraform/services/api-waf-sync/main.tf
jdettmannnava: We don't use it any more

carlpartridge: Neither our lambda nor DPCs seemed to need it so we got rid of it.  Both our lambdas are using ~ `ipSetName := fmt.Sprintf("bcda-%s-api-customers", os.Getenv("ENV"))`.

terraform/services/api-waf-sync/main.tf
carlpartridge: DPC has a bit more complicated a route into their DB then BCDA seems to and our parameter stores have different existing variables.  Could be nice to simplify that at some point?

carlpartridge:
Im currently seeing this error:
```
╷
│ Error: no matching EC2 Security Group found
│ 
│   with data.aws_security_group.db,
│   on main.tf line 34, in data "aws_security_group" "db":
│   34: data "aws_security_group" "db" {
│ 
╵
```
And not entirely sure how best to resolve it.  Im not seeing anything that seems to match that for BCDA.

gfreeman-navapbc:
> Im currently seeing this error:
> 
> ```
> ╷
> │ Error: no matching EC2 Security Group found
> │ 
> │   with data.aws_security_group.db,
> │   on main.tf line 34, in data "aws_security_group" "db":
> │   34: data "aws_security_group" "db" {
> │ 
> ╵
> ```
> 
> And not entirely sure how best to resolve it. Im not seeing anything that seems to match that for BCDA.

I think the BCDA database security groups are named `bcda-[env]-rds` but I may be mistaken

carlpartridge:
> I think the BCDA database security groups are named `bcda-[env]-rds` but I may be mistaken

Hmm it didnt seem to like my changes:
```
locals {
  full_name    = "${var.app}-${var.env}-api-waf-sync"
  bcda_sg_name = "bcda-${var.env}-rds"
  dpc_sg_name  = "dpc-${var.env}-db"
}

data "aws_security_group" "db" {
  name = "local.${var.app}_sg_name"
}
```
Whats the better way to do this?

terraform/services/api-waf-sync/main.tf
gsf: Yeah okay I'm fine with punting on it for now.

terraform/services/api-waf-sync/main.tf
carlpartridge: @gsf @gfreeman-navapbc  It also occurs to me that our DB_URL contains all the DB info, user, pw, host, instance, etc, is that problematic?

gfreeman-navapbc:
Something similar to the `locals` portion of the api-waf service itself: https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/0b23d0fa5e6978eae850ae88c1fe2f284e2a8f50/terraform/services/api-waf/main.tf#L1-L17

terraform/services/api-waf-sync/main.tf
gsf: Good point to raise. Terraform should mark the data as sensitive but to ensure the value is redacted in the logs it's better to accept the value as a variable and retrieve it in the workflow with aws-params-env-action. See an example of that at https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/0b23d0fa5e6978eae850ae88c1fe2f284e2a8f50/.github/workflows/api-rds-plan.yml#L40

carlpartridge: @gsf Ok, let me know if how I implemented this in the intended way.  Also one question this brings up is that (as far as I can tell) the terraform plan/apply happens only in dev env, which would grab the dev version of the parameter store, and pass that through to tf?  Which sounds problematic when it comes to other envs?

gsf: What you've done looks correct, though the overlapping DB_HOST and DATABASE_URL variables are still making me itch.

Another way this could be managed (in both BCDA and DPC lambdas) would be to grab the AWS param for the database URL at runtime, similar to how the bucket role ARN is retrieved in the optout lambda here: https://github.com/CMSgov/bcda-app/blob/b8ae6d58d4379111a8af3d3eb584e8a80d3aad78/bcda/lambda/optout/main.go#L79

Looks like you already have a helper function to do this even: https://github.com/CMSgov/bcda-app/blob/b8ae6d58d4379111a8af3d3eb584e8a80d3aad78/bcda/database/config.go#L60

Pulling the parameter value directly into the lambda avoids passing it through CI/CD and terraform entirely. The downside of the runtime retrieval from the parameter store is that it would increase the number of calls, but given the likely frequency at which this lambda will run I don't see that as an issue here.

jdettmannnava: Would it be possible to build a map in locals, so
db_param={"BCDA": ..., "DPC": ...} and call aws ssm with that?

gsf: @jdettmannnava It would be better for the DPC lambda to likewise get the value from parameter store directly instead of relying on an environment variable.

gsf (approved):

PR #158, opened 2025-01-02
PLT-832: Update upload-artifacts from v3 to v4

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-832

## 🛠 Changes

Changed upload-artifacts version to v4.

## ℹ️ Context

The current upload-artifacts version is being sunset, so we will have to update the version to v4.

## 🧪 Validation

Workflow should run without failures

### DIFF
--- a/.github/workflows/aws-params-env-action-check-dist.yml
+++ b/.github/workflows/aws-params-env-action-check-dist.yml
@@ -46,7 +46,7 @@ jobs:
         id: diff
 
       # If index.js was different than expected, upload the expected version as an artifact
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         if: ${{ failure() && steps.diff.conclusion == 'failure' }}
         with:
           name: dist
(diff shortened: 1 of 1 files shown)


gsf (commented):
Looks good. You could temporarily (or permanently) add .github/workflows/aws-params-env-action-check-dist.yml to the paths in the pull_request trigger of this workflow to check it in this PR.

gfreeman-navapbc:
> Looks good. You could temporarily (or permanently) add .github/workflows/aws-params-env-action-check-dist.yml to the paths in the pull_request trigger of this workflow to check it in this PR.

@gsf looks like it succeeds, but the workflow doesn't actually run the upload since the diff doesn't produce anything.

gsf (approved):
We could go so far as to comment out the "if" clause but seeing as we're not getting into multiple artifacts or hidden files as mentioned in [the migration notes](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md), I think this is good to go.

PR #165, opened 2025-01-17
PLT-864: ADD IPv6 IP-set for BCDA non-SBX environments

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-864

## 🛠 Changes
Adds a new IP set for BCDA environments and cleans up some repeated conditionals with a variable

## ℹ️ Context

BCDA team was running into issues deploying their new WAF sync lambda due to our IP set not allowing IPv6 addresses.

[Relevant slack thread.](https://cmsgov.slack.com/archives/C04UG13JF9B/p1737139629675179)

## 🧪 Validation
BCDA should be able to add IPv6 addresses to this IP set after release.

### DIFF
--- a/terraform/services/api-waf/main.tf
+++ b/terraform/services/api-waf/main.tf
@@ -41,6 +41,7 @@ resource "aws_wafv2_ip_set" "api_customers" {
 }
 
 resource "aws_wafv2_ip_set" "ipv6_api_customers" {
+  count              = local.is_sandbox ? 0 : 1
   name               = "${var.app}-${var.env}-ipv6-api-customers"
   description        = "IP ranges for customers of this API"
   scope              = "REGIONAL"


carlpartridge (approved):

terraform/services/api-waf/main.tf
```diff
@@ -41,6 +42,25 @@ resource "aws_wafv2_ip_set" "api_customers" {
...
+  count              = var.app == "bcda" ? 1 : 0
+  name               = "${var.app}-${var.env}-ipv6-api-customers"
+  description        = "IP ranges for customers of this API"
+  scope              = "REGIONAL"
+  ip_address_version = "IPV6"
+
+  # Addresses will be managed outside of terraform. This is
+  # a placeholder address for all apps/environments.
+  # See: https://confluence.cms.gov/x/UDs2Q
+  addresses = []
```
gsf, line 55: Can this be applied with an empty array? Possibly! I'm seeing now that [the docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_ip_set) say "Contains an array of strings that specifies zero or more IP addresses or blocks of IP addresses."

If so, would you remove the "This is a placeholder address for all apps/environments." sentence from the comment? And update the addresses attribute in api_customers to match?

terraform/services/api-waf/main.tf
```diff
@@ -41,6 +42,25 @@ resource "aws_wafv2_ip_set" "api_customers" {
   }
 }
 
+resource "aws_wafv2_ip_set" "ipv6_api_customers" {
+  count              = var.app == "bcda" ? 1 : 0
```
gsf, line 46: Any reason not to include an empty ipv6 IP set for all of the APIs?

gfreeman-navapbc: Not really, I can remove this constraint

gsf (approved):

PR #168, opened 2025-01-27
PLT-884 Increase compute for GHA runners

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-884

## 🛠 Changes

Changes instance size for runners to t3.xlarge

## ℹ️ Context

Application engineers expect self-hosted runners to match github-hosted runners for compute.

## 🧪 Validation

See terraform plan in checks, and see apply on merge to ensure successful infra update.

### DIFF
--- a/packer/github-actions-runner/variables.pkr.hcl
+++ b/packer/github-actions-runner/variables.pkr.hcl
@@ -7,7 +7,7 @@ variable "region" {
 variable "instance_type" {
   description = "The instance type Packer will use for the builder"
   type        = string
-  default     = "t3.xlarge"
+  default     = "t3.large"
 }
 
 variable "ami_account" {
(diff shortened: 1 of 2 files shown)


packer/github-actions-runner/variables.pkr.hcl
```diff
@@ -7,7 +7,7 @@ variable "region" {
 variable "instance_type" {
   description = "The instance type Packer will use for the builder"
   type        = string
-  default     = "t3.large"
+  default     = "t3.xlarge"
```
gsf, line 10: This change can be dropped. The size that packer uses for building the image doesn't affect the size of the instances used for running workflows.

gsf (approved):

PR #169, opened 2025-01-28
PLT-884: Add config for block storage for GHA runner AMI

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-884

## 🛠 Changes

Specifies volume size for GHA runner AMIs

## ℹ️ Context

We are still not utilizing the full disk space of the runners. This change will double the spacce and potentially unblock the jobs that are seeing disk usage errors.

## 🧪 Validation

Workflows utilizing the GHA runners should no longer fail due to storage errors.


gsf (approved):
Looks good! We'll have to check on instances created to ensure the device name matches the root volume.

PR #170, opened 2025-01-29
PLT-785: Add AB2D Sandbox to WAF environments

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-785

## 🛠 Changes

Adds AB2D sandbox to the list of environments we deploy the shared WAF terraform to.

## ℹ️ Context

We want to test the changes in all environments before we rollout to production. The new shared terraform will bring all the teams to one consistent state 

## 🧪 Validation

AB2D sandbox access should remain as usual.


.github/workflows/api-waf-apply.yml
```diff
@@ -22,15 +22,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        app: [bcda, dpc]
+        app: [ab2d, bcda, dpc]
         env: [dev, test, sbx]
```
juliareynolds-nava, line 26: Only dpc has a prod environment, is that correct?  Otherwise I would expect env: [dev, test, sbx, prod].

.github/workflows/api-waf-apply.yml
```diff
@@ -23,12 +23,14 @@ jobs:
       fail-fast: false
       matrix:
         app: [bcda, dpc]
```
juliareynolds-nava, line 25: Should ab2d be included in the app list?

gfreeman-navapbc: No, I'll add ab2d as a part of the prod rollout since it'd be more of a code change to exclude ab2d prod than just adding the appropriate variables into the environment

PR #171, opened 2025-01-29
PLT-895: Increase Concurrent GHA Runners

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-895

## 🛠 Changes

Overrides the default 3 concurrent runners to 12 to allow for long-running jobs to not block runners being assigned.

## ℹ️ Context

See relevant slack thread: https://cmsgov.slack.com/archives/C04UG13JF9B/p1738176789390989?thread_ts=1737999040.128209&cid=C04UG13JF9B

This should allow for workflows with multiple jobs to be run in parallel without blocking.

## 🧪 Validation

Teams should see their runners pick up more jobs in workflows.


gsf (approved):
We could also look into bumping up the idleCount as well to make it so more subsequent jobs start quickly, but this is a good change for now.

PR #177, opened 2025-02-06
PLT-897: Add new shell provisioner for resizing volumes

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-897

## 🛠 Changes

Adds a new provisioner to the packer build.

## ℹ️ Context

We need to actually resize the volumes for the jobs to have enough space on disk to succeed. We're embedding resizing of the volumes into packer.

## 🧪 Validation

Jobs should get past the disk space issues.

### DIFF
--- a/packer/github-actions-runner/build.pkr.hcl
+++ b/packer/github-actions-runner/build.pkr.hcl
@@ -25,6 +25,18 @@ build {
     ]
   }
 
+  provisioner "shell" {
+    remote_folder = "/home/ec2-user/"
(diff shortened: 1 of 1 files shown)


gsf (approved):
Looks good! Make sure you run https://github.com/CMSgov/ab2d-bcda-dpc-platform/actions/workflows/github-actions-runner-apply.yml after merging to point the launch template at the latest image.

PR #178, opened 2025-02-10
PLT-905: Add Docker Compose install to packer build for GHA runners

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-905

## 🛠 Changes

Adds installation commands for the latest release of docker compose for the GHA runners packer build.

## ℹ️ Context

Several teams use containerized tests in their workflows, this will allow them to run without requiring the teams to specify a docker compose installation in every workflow.

## 🧪 Validation

Compose should be installed by default on all runner images after this is merged and applied.

### DIFF
--- a/packer/github-actions-runner/build.pkr.hcl
+++ b/packer/github-actions-runner/build.pkr.hcl
@@ -17,8 +17,9 @@ build {
     remote_folder = "/home/ec2-user/"
     inline = [
       "sudo dnf install -y amazon-cloudwatch-agent jq git docker libicu curl",
-      "sudo curl -L https://github.com/docker/compose/releases/latest/download/docker-compose-$(uname -s)-$(uname -m) -o /usr/local/bin/docker-compose",
-      "sudo chmod +x /usr/local/bin/docker-compose",
+      "mkdir -p /usr/local/lib/docker/cli-plugins"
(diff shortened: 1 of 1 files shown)


carlpartridge (approved):

gsf (commented):
Hmm why are we not installing the Docker Compose plugin as documented at https://docs.docker.com/compose/install/linux/#install-the-plugin-manually? I think of "docker compose" as being the current way to run it.

gfreeman-navapbc:
> Hmm why are we not installing the Docker Compose plugin as documented at https://docs.docker.com/compose/install/linux/#install-the-plugin-manually? I think of "docker compose" as being the current way to run it.

Good question, must have found some old documentation. Will update.

packer/github-actions-runner/build.pkr.hcl
```diff
@@ -17,6 +17,9 @@ build {
     remote_folder = "/home/ec2-user/"
     inline = [
       "sudo dnf install -y amazon-cloudwatch-agent jq git docker libicu curl",
+      "mkdir -p /usr/local/lib/docker/cli-plugins"
```
gsf, line 20:
```suggestion
      "mkdir -p /usr/local/lib/docker/cli-plugins",
```

juliareynolds-nava (approved):

PR #186, opened 2025-03-04
PLT-508: AB2D WAF IP Set Sync Workflow

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-508

## 🛠 Changes

Adds automation in the form of a workflow and associated script for automating AB2D WAF IP set updates.

## ℹ️ Context

AB2D needs an automated way to keep their IP sets up to date, this aims to do that. This workflow will eventually run on a cron schedule the team sees fit (likely daily/multiple times a day). The method is as follows (modeled off of the steps found [here](https://repost.aws/knowledge-center/maintain-waf-ipsets-using-cli)):

1. IPs are stored in the `cdap-private` repo.
2. Workflow is triggered by cron.
3. IPs are loaded into JSON that is digestible by AWS `update-ip-set` command. ([docs](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/wafv2/update-ip-set.html))
4. Existing IP sets are fetched with `get-ip-set` to get a lock token needed for the update command. ([docs](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/wafv2/get-ip-set.html))
5. Update command is then run with the lock token provided by the previous step.

## 🧪 Validation

Once merged, this workflow will automatically update the IP sets on a given schedule for all environments specified in the workflow file. (dev, test, prod)

### DIFF
--- a/.github/workflows/ab2d-ip-sets-sync.yml
+++ b/.github/workflows/ab2d-ip-sets-sync.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         app: [ab2d]
-        env: [dev, test, prod]
+        env: [dev, test]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/checkout@v4
--- a/.github/workflows/ab2d-ip-sets-sync.yml
+++ b/.github/workflows/ab2d-ip-sets-sync.yml
@@ -1,0 +1,36 @@
+name: AB2D WAF IP Sets Sync
+
+on:
+  workflow_dispatch: # Allow manual trigger
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  waf-ip-sync:
+    runs-on: self-hosted
+    defaults:
+      run:
+        working-directory: ./scripts
+    strategy:
+      matrix:
+        app: [ab2d]
+        env: [dev, test, prod]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
+        with:
+          repository: CMSgov/cdap-private
+          ref: 'main'
+          sparse-checkout: ip_sets/ab2d/allowed_ips.txt
+          sparse-checkout-cone-mode: false
+          token: ${{ env.GITHUB_TOKEN }}
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::${{ secrets.BCDA_ACCOUNT }}:role/delegatedadmin/developer/bcda-mgmt-github-actions
+          aws-region: ${{ vars.AWS_REGION }}
+      - run: bash waf-ip-set-sync.sh
+        env:
+          APP: ${{ matrix.app }}
+          ENV: ${{ matrix.env }}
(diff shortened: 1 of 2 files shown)


scripts/waf-ip-set-sync.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "Generating IP set"
+
+IPV4_LIST=$(grep -v '^#' allowed_ips.txt | jq -Rs '{Addresses: split("\n") | map(select(length > 0))}')
+
+echo "Fetching IP set IDs"
+
+{
+  read -r IPV4_SET_ID
+  read -r IPV6_SET_ID
```
juliareynolds-nava, line 13: Does this IPV6 variable exist?

juliareynolds-nava (approved):

scripts/waf-ip-set-sync.sh
gfreeman-navapbc: Yes, we create both of the IP sets in the WAF terraform here: https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/b93e40fa10e5b844e75aecf1d654f9505268164e/terraform/services/api-waf/main.tf#L26-L58

.github/workflows/ab2d-ip-sets-sync.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/checkout@v4
+        with:
+          repository: CMSgov/cdap-private
+          ref: 'main'
+          sparse-checkout: ip_sets/ab2d/allowed_ips.txt
+          sparse-checkout-cone-mode: false
+          token: ${{ env.GITHUB_TOKEN }}
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::${{ secrets.BCDA_ACCOUNT }}:role/delegatedadmin/developer/bcda-mgmt-github-actions
```
gsf, line 31: Need to assume a role in the account where the WAF is, correct?

scripts/waf-ip-set-sync.sh
```diff
@@ -0,0 +1,26 @@
...
+
+LOCK_TOKEN=$(aws wafv2 get-ip-set --name $APP-$ENV-api-customers --scope REGIONAL --id $IPV4_SET_ID --region us-east-1 | jq '.LockToken')
+
+aws wafv2 update-ip-set \
+  --name $APP-$ENV-api-customers \
+  --scope REGIONAL \
+  --id $IPV4_SET_ID \
+  --region us-east-1 \
+  --addresses $IPV4_LIST \
+  --lock-token $LOCK_TOKEN
```
gsf, line 26: Based on the note at https://docs.aws.amazon.com/waf/latest/APIReference/API_UpdateIPSet.html it appears that update-ip-set will replace all existing addresses in the set, which is what we want, but we should confirm that behavior.

scripts/waf-ip-set-sync.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+set -euo pipefail
+
+echo "Generating IP set"
+
+IPV4_LIST=$(grep -v '^#' allowed_ips.txt | jq -Rs '{Addresses: split("\n") | map(select(length > 0))}')
```
gsf, line 7: Nice use of jq

PR #189, opened 2025-03-07
PLT-948: Add clone to temp directory

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-948

## 🛠 Changes

Changes the git clone method to one where we clone to a temp directory to get around git clone's limitation of needing the directory to be empty

## ℹ️ Context

After implementing a fix where we check to see if tfenv is already installed, we were still getting errors with `git clone` because the directory already existed.

## 🧪 Validation

Workflows should pass and not be stopped by previous tfenv installations.

### DIFF
--- a/actions/setup-tfenv-terraform/action.yml
+++ b/actions/setup-tfenv-terraform/action.yml
@@ -14,7 +14,6 @@ runs:
         else
           git clone --depth=1 https://github.com/tfutils/tfenv.git temp
           mv temp ~/.tfenv
-          rm -rf temp
           echo "PATH=$HOME/.tfenv/bin:$PATH" >> "$GITHUB_ENV"
         fi
       shell: bash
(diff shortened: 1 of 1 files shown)


actions/setup-tfenv-terraform/action.yml
```diff
@@ -12,7 +12,9 @@ runs:
         if tfenv --version; then
           echo "tfenv already installed, skipping"
         else
-          git clone --depth=1 https://github.com/tfutils/tfenv.git ~/.tfenv
+          git clone --depth=1 https://github.com/tfutils/tfenv.git temp
+          mv temp ~/.tfenv
+          rm -rf temp
```
gsf, line 17: Hmm now that I look at this (and the answer at https://stackoverflow.com/a/13852329/589391), I can think of no need for the `rm -rf temp` line since the `mv` will remove that directory.

gsf: And, ugh, sorry to add more to this, but to make sure we avoid any conflicts in the directory where this is running, we should really do this:

```
TMPDIR=$(mktemp -d)
git clone --depth=1 https://github.com/tfutils/tfenv.git $TMPDIR
mv $TMPDIR ~/.tfenv
```

PR #196, opened 2025-03-20
PLT-506: External services WAF IP sync

jscott-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-506

## 🛠 Changes

- Added new file external-services-ip-sets-sync.yml runner file
- Added new file waf-ip-set-sync-external-services.sh to run the sync via the AWS CLI.
- Added new file allowed-ips.txt under cdap-private/ip-sets/external-services with IP sets taken from BCDA account.

## ℹ️ Context

These changes are intended to establish a framework for syncing the external-services IPSets from various accounts and environments to a file of CIDR IP ranges in the CDAP private repo. This workflow is currently configured to be manually run, but can easily be reconfigured to run on a schedule or on an update to the CDAP private repo.

## 🧪 Validation

These changes were validated by running a GHA workflow against the AB2D IMPL and BCDA/DPC accounts us-east-1 regions.

### DIFF
--- a/.github/workflows/external-services-ip-sets-sync.yml
+++ b/.github/workflows/external-services-ip-sets-sync.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
-          role-to-assume: arn:aws:iam::${{ secrets[format('{0}_{1}_ACCOUNT', matrix.app, matrix.env)] }}:role/delegatedadmin/developer/${{ matrix.app }}-${{ matrix.env }}-github-actions
+          role-to-assume: arn:aws:iam::${{ matrix.app == 'ab2d' && secrets[format('{0}_{1}_ACCOUNT', matrix.app, matrix.env)] || secrets.BCDA_ACCOUNT }}:role/delegatedadmin/developer/${{ matrix.app }}-${{ matrix.env }}-github-actions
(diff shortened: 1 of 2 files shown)


.github/workflows/external-services-ip-sets-sync.yml
```diff
@@ -0,0 +1,50 @@
+name: External Services WAF IP Sets Sync
+
+on:
+  workflow_dispatch: # Allow manual trigger
```
gfreeman-navapbc, line 4: I think you could add this just for testing, so we don't have to merge this to main before we see if it fails for one reason or another while executing on the GHA runner. Once we have a successful build we can take it out and merge to main. Did we ever specify a schedule this should run on as well? Or will we be manually running this like the AB2D IP set sync?

```suggestion
on:
  workflow_dispatch: # Allow manual trigger
  push:
    branches:
      - jscott/PLT-506
```

juliareynolds-nava (approved):

.github/workflows/external-services-ip-sets-sync.yml
gfreeman-navapbc: And just now seeing the runs from this in test, so disregard this suggestion

jscott-nava:
This PR has now been fully tested against all environments and has synced them all to the same 47 shared external services IP ranges.

gfreeman-navapbc (approved):

PR #199, opened 2025-04-01
[PLT-890] added bucket and backend config

juliareynolds-nava:
## 🎫 Ticket

## 🛠 Changes

Added s3 bucket and backend config to vpc main

## ℹ️ Context

Add backends to the TF in the platform repo for Greenfield migration.

## 🧪 Validation

running terraform apply created 3s buckets in each vpc

### DIFF
--- a/terraform/modules/vpc/main.tf
+++ b/terraform/modules/vpc/main.tf
@@ -20,3 +20,16 @@ data "aws_vpc" "this" {
     }
   }
 }
+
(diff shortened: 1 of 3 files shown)


gsf (commented):
So yep as discussed I think best to use this PR to just fix the DB_HOST quoting issue, then open another PR for the update to the bucket module, then one for the tfstate service.

PR #204, opened 2025-04-07
[PLT-890] Add a backend bucket to each vpc upon creation

juliareynolds-nava:
## 🎫 Ticket

## 🛠 Changes

created a bucket called "backend-bucket" in each vpc.

## ℹ️ Context

So that terraform state can be managed for vpcs

## 🧪 Validation

PR checks pass


gsf (commented):
I think this PR is unnecessary. The tfstate bucket is created and managed by the terraform in terraform/services/tfstate/main.tf.

PR #205, opened 2025-04-07
[PLT-890] add backend bucket to tfstate service

juliareynolds-nava:
## 🎫 Ticket

## 🛠 Changes

Added backend bucket to the tfstate service
## ℹ️ Context
So that the tfstate service can manage state.

## 🧪 Validation
PR checks


gsf (commented):
The bucket is already created in terraform/services/tfstate/main.tf. What's needed is an update to terraform/modules/bucket/main.tf to modify the bucket resource like so:

```
resource "aws_s3_bucket" "this" {
  bucket        = var.legacy == true ? var.name : null
  bucket_prefix = var.legacy == false ? var.name : null
  force_destroy = true
}
```

Then, once that's been run from local and tested, the tfstate plan and apply workflows must be modified to pass in the "legacy" variable.

terraform/services/tfstate/terraform.tf
```diff
@@ -7,10 +7,17 @@ provider "aws" {
...
   backend "s3" {
-    key = "tfstate/terraform.tfstate"
+    key            = "tfstate/terraform.tfstate"
+    bucket         = ""
+    dynamodb_table = ""
+    encrypt        = true
+
   }
-}
+}
```
gsf, line 23: There should be no changes to this file in the PR.

PR #223, opened 2025-05-05
PLT-1040: Add bucket SSM Parameter back to Greenfield Environments

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1040

## 🛠 Changes

This adds back an optional parameter that references the name of the bucket passed in through the `function` module. 

## ℹ️ Context

We removed the SSM Parameter for the bucket since it was creating parameters in every environment on legacy as well as greenfield. This removal makes greenfield development a bit harder as we cannot interpolate the bucket name by by App and Environment during deployment.

## 🧪 Validation

This should allow DPC to continue greenfield development and deployment

### DIFF
--- a/terraform/modules/function/main.tf
+++ b/terraform/modules/function/main.tf
@@ -146,7 +146,8 @@ module "zip_bucket" {
     "arn:aws:iam::${data.aws_ssm_parameter.sbx_account[0].value}:role/delegatedadmin/developer/${var.app}-sbx-github-actions",
   ] : []
 
-  legacy = var.legacy
+  legacy        = var.legacy
+  ssm_parameter = var.legacy ? null : "/${var.app}/${var.env}/${var.name}-bucket"
 }
 
 resource "aws_s3_object" "empty_function_zip" {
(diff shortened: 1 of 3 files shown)


terraform/modules/function/main.tf
```diff
@@ -146,7 +146,8 @@ module "zip_bucket" {
     "arn:aws:iam::${data.aws_ssm_parameter.sbx_account[0].value}:role/delegatedadmin/developer/${var.app}-sbx-github-actions",
   ] : []
 
-  legacy = var.legacy
+  legacy        = var.legacy
```
jscott-nava, line 149: Is 'legacy' used for anything?

mjburling: Because it defaults to `true`, we can preserve the legacy behavior without any changes to the existing workflows, and only trigger greenfield behavior by explicitly setting it to `false`, e.g. `-var=legacy=false`.

gsf (approved):

PR #226, opened 2025-05-13
DPC-4699: Support Opt-Out, WAF Sync GF Deployments

mjburling:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4699

## 🛠 Changes
This PR is primarily concerned with updates that enable the Opt-Out Import and Opt-Out Export Lambda deployments for DPC, however, it introduces some other minor fixes along the way including:
- ~relaxes~ removes the constraint on an effectively pinned aws provider version 5.8.0 that was release ca July 2023
- removes the terraform lock files that should have been ignored from the outset
- extends rights of the DASG Application Admins to enjoy some of the same trust as the specific ADO application admins
- foregoes rds instance address lookup in parameter store in favor of rds instance lookups based on application/environment tags for greenfield–existing legacy environments continue to use legacy SSM parameters.

## 🧪 Validation

This was manually applied in greenfield against DPC. Plans have been successful across the legacy environments, with spot checks for the updated values.


gfreeman-navapbc (approved):

terraform/modules/function/main.tf
```diff
@@ -21,6 +21,10 @@ data "aws_iam_role" "admin" {
   name = var.app == "dpc" ? "ct-ado-bcda-application-admin" : "ct-ado-${var.app}-application-admin"
 }
 
+data "aws_iam_role" "dasg_admin" {
+  name = "ct-ado-dasg-application-admin"
+}
+
```
gsf, line 27: Where is this role found? We use ct-ado-bcda-application-admin in the greenfield accounts.

terraform/modules/function/versions.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 5.8.0"
+      version = "~> 5"
     }
   }
 }
```
gsf, line 8: These `required_providers` blocks can be dropped.

terraform/modules/function/main.tf
mjburling: Sure. It's for federal staff on DASG–I'm not part of the `ct-ado-bcda-application-admin` 😉

gsf: Okay then can we add the bcda admin as well? It only serves to aid in troubleshooting and doesn't serve us well if we can't assume the role.

mjburling: Yes, exactly! That's why I've added the DASG admin here. The content of `data.aws_iam_role.admin.arn` defined just above will need to be adjusted for greenfield in a subsequent change set ahead of following through for `ab2d`, but `ct-ado-bcda-application-admin` is already there for gf-dev `dpc` and `bcda` resources.

gsf (approved):

PR #228, opened 2025-05-15
PLT-805 Sync sg rules and IP sets

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-805

## 🛠 Changes

Added workflows for syncing security group rules and IP sets in greenfield. Also updated api-waf terraform for greenfield.

## ℹ️ Context

This is coordinated with https://github.com/CMSgov/cdap-private/pull/6 for updated scripts and CIDR lists used in syncing security groups and IP sets. 

## 🧪 Validation

Terraform applied and scripts tested locally. We can expect api-waf plan and apply to fail in greenfield environments where load balancers have not yet been created.

The sync-ab2d-ip-set workflow failed as expected in https://github.com/CMSgov/ab2d-bcda-dpc-platform/actions/runs/15053120350 since there's no WAF or IP sets created yet in greenfield for AB2D.

The sync-external-services-ip-sets workflow at https://github.com/CMSgov/ab2d-bcda-dpc-platform/actions/runs/15056583885 succeeds.


gfreeman-navapbc (approved):
I like the plan & apply workflow files, could probably be replicated across most our workflows

PR #231, opened 2025-05-19
PLT-1033: Introduce a platform child module

mjburling:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1033

## 🛠 Changes

Introduces a child module for commonly referenced security groups, iam roles, subnets, and other basics that any consuming module might otherwise lookup for itself with a separate collection of data sources and input parameters. 😬 

## ℹ️ Context

This will help define the standards by which the CDAP customers operate, including setting a baseline for default tags.

## 🧪 Validation

Because this is for greenfield only, I've been able to make use of this child module in a limited capacity in the AB2D core service module–it works!

### DIFF
--- a/terraform/modules/platform/README.md
+++ b/terraform/modules/platform/README.md
@@ -1,6 +1,32 @@
 # Platform Child Module
 
-This is a simple child module comprised of data sources, outputs, and some modest logic to encourage adoption of emerging, platform-wide standards.
+This simple [child module](https://developer.hashicorp.com/terraform/language/modules#child-modules) comprises data sources, outputs, and some modest logic to encourage adoption of DASG's emerging, _platform_-wide standards for use in CDAP-customer terraform modules.
+The resources that are referenced by terraform data source in this module are managed by the CMS Hybrid Cloud team and/or the CDAP team.
+
+## Example Usage
+
+```hcl
+## AB2D API Module Usage Example
+module "platform" {
+  # Ensure `ref` in the following line is pinned to something static
+  # e.g. a known branch, commit hash, or tag from **this repository**
+  source = "git::https://github.com/CMSgov/ab2d-bcda-dpc-platform.git//terraform/modules/platform?ref=plt-1033"
+
+  app         = "ab2d"
+  env         = var.env
+  root_module = "https://github.com/CMSgov/ab2d-ops/tree/main/terraform/services/api"
+  service     = "api"
+}
+
+## Configure the aws provider with the default tag standards sourced from the `platform` child module
+provider "aws" {
+  region = "us-east-1"
+
+  default_tags {
+    tags = module.platform.default_tags
+  }
+}
+```
 
 <!-- BEGIN_TF_DOCS -->
 <!--WARNING: GENERATED CONTENT with terraform-docs, e.g.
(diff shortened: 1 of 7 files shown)


oluwolenpbc (approved):

juliareynolds-nava (approved):

terraform/modules/platform/variables.tf
```diff
@@ -0,0 +1,33 @@
+variable "app" {
+  description = "The short name for the delivery team or ADO."
+  type        = string
+  validation {
+    condition     = contains(["ab2d", "bcda", "dpc"], var.app)
+    error_message = "Invalid short var.app (application). Must be one of ab2d, bcda, or dpc."
```
gsf, line 6: We'll need "cdap" for resources for the shared codebuild runners unless we use "bcda" as the broader application name.

terraform/modules/platform/variables.tf
```diff
@@ -0,0 +1,33 @@
...
+variable "additional_tags" {
+  default     = {}
+  description = "Additional tags to merge into final default_tags output"
+  type        = map(string)
+}
+
+variable "module_root" {
+  description = "The full URL to the terraform module root at issue for this infrastructure"
+  type        = string
+}
```
gsf, line 33: Do you prefer module_root over root_module? I see it referred to as "root module" in the terraform docs.

mjburling: I'm ambivalent.

terraform/modules/platform/variables.tf
mjburling: I'll meditate on this. I was really unhappy with having to commit to counts on a subset of data sources that result from cdap's support for networking just in production. Maybe there's a straightforward solution that isn't terrible.

gsf: We can leave it at the three apps for now and make the decision later.

terraform/modules/platform/README.md
```diff
@@ -0,0 +1,93 @@
+# Platform Child Module
+
+This is a simple child module comprised of data sources, outputs, and some modest logic to encourage adoption of emerging, platform-wide standards.
```
gsf, line 3:
```suggestion
This is a simple child module comprising data sources, outputs, and some modest logic to encourage adoption of emerging, platform-wide standards.
```

gsf (approved):

gsf:
As I mentioned in Slack, I think calling the module "standards" would get across the point of it better than the more ambiguous "platform", but I can live with it.

terraform/modules/platform/variables.tf
mjburling: I think I'm prepared to make the decision now–this module should **not** be used by cdap. We ought to keep this as simple as possible. To that end, I think if cdap resources need to apply similar standards, we should adopt a module specific to those needs.

Further, as tempting as it is, I would urge that we avoid trying our hand at an inheritance strategy via a third module that focuses on one standard, e.g. `default_tags`. If there are two places that need updates for some standard, it needs to be well understood that there are exactly two places that serve as the two entry points to accommodate those future enhancements.

gsf: Oh, so this module should not be used by any services in this repo?

PR #234, opened 2025-05-21
PLT-1031 Update for clean apply in dpc-test

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1031

## 🛠 Changes

Updates made to api-rds to apply in greenfield dpc-test environment.

## ℹ️ Context

This PR minimizes the resources created in greenfield around a restored database.

## 🧪 Validation

Fully applied on the dpc-test environment in greenfield. See checks for plans in other environments.

### DIFF
--- a/terraform/services/api-rds/main.tf
+++ b/terraform/services/api-rds/main.tf
@@ -108,17 +108,17 @@ resource "aws_vpc_security_group_ingress_rule" "db_access_from_controller" {
 }
 
 resource "aws_vpc_security_group_ingress_rule" "db_access_from_mgmt" {
-  count             = var.app == "ab2d" || var.app == "dpc" ? 1 : 0
+  count             = var.legacy && (var.app == "ab2d" || var.app == "dpc") ? 1 : 0
   description       = "Management VPC Access"
   from_port         = 5432
   to_port           = 5432
   ip_protocol       = "tcp"
-  cidr_ipv4         = var.legacy ? var.mgmt_vpc_cidr : data.aws_ssm_parameter.cdap_mgmt_vpc_cidr[0].value
+  cidr_ipv4         = var.mgmt_vpc_cidr
   security_group_id = aws_security_group.sg_database.id
 }
 
 resource "aws_vpc_security_group_ingress_rule" "additional_ingress" {
-  for_each                     = var.app == "bcda" || var.app == "dpc" ? toset(local.additional_ingress_sgs) : toset([])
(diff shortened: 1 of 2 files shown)


juliareynolds-nava (approved):

terraform/services/api-rds/main.tf
```diff
@@ -81,16 +81,17 @@ locals {
...
-  vpc_id = module.vpc.id
+  vpc_id = var.legacy ? module.vpc[0].id : module.platform[0].vpc_id
 
-  tags = merge(
+  tags = var.legacy ? merge(
     data.aws_default_tags.data_tags.tags,
     { "Name" = local.sg_name },
     var.app == "dpc" ? local.dpc_specific_tags : {}
-  )
+  ) : {}
```
mjburling, line 94: I think the tag Name is traditional for security groups. I think you're going to want those everywhere.

Also, you probably ought to drop `data.aws_default_tags.data_tags` all together–I don't think there's a single legitimate use for them in this module.

terraform/services/api-rds/main.tf
```diff
@@ -281,7 +288,7 @@ resource "aws_db_instance" "api" {
       var.app == "dpc" && var.env == "sbx") ? "4HR Daily Weekly Monthly" : "Daily Weekly Monthly"
     },
     var.app == "dpc" ? local.dpc_specific_tags : {}
-  )
+  ) : {}
```
mjburling, line 291: Like elsewhere, the reference to `data.default_tags.data_tags` should be removed along with the definition for the data source itself. Dropping `Name` from the greenfield is a good idea 👍 . I'm not actually sure if it's useful in legacy, either, but... that can come another day 😉

.github/workflows/tf-api-rds.yml
```diff
@@ -0,0 +1,57 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        app: [ab2d, bcda, dpc]
+        env: [dev]
+        include:
+          - app: bcda
+            env: test
+          - app: dpc
+            env: test
```
mjburling, line 42: Please exclude `ab2d` for the time being.
```suggestion
        app: [bcda, dpc]
        env: [dev, test]
```

mjburling (changes requested):

.github/workflows/tf-api-rds.yml
gsf: Why? We're already managing the ab2d-dev database infra with this service, and plan shows no changes.

mjburling (approved):

PR #242, opened 2025-05-29
PLT-1091 Allow sandbox env in functions

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1091

## 🛠 Changes

Add the sandbox env in the function module and BCDA admin services.

## ℹ️ Context

Another follow-up on #238 to allow for apply of terraform for BCDA admin functions in the sandbox environment.

## 🧪 Validation

See checks, specifically around sandbox envs.

### DIFF
--- a/terraform/services/admin-create-aco/variables.tf
+++ b/terraform/services/admin-create-aco/variables.tf
@@ -8,11 +8,11 @@ variable "app" {
 }
 
 variable "env" {
-  description = "The application environment (dev, test, sbx, prod)"
+  description = "The application environment (dev, test, sbx, sandbox, prod)"
   type        = string
   validation {
-    condition     = contains(["dev", "test", "sbx", "prod"], var.env)
-    error_message = "Valid value for env is dev, test, sbx or prod."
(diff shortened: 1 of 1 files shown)


terraform/modules/function/variables.tf
```diff
@@ -8,11 +8,11 @@ variable "app" {
 }
 
 variable "env" {
-  description = "The application environment (dev, test, sbx, prod)"
+  description = "The application environment (dev, test, sbx, sandbox, prod)"
   type        = string
   validation {
-    condition     = contains(["dev", "test", "sbx", "prod"], var.env)
-    error_message = "Valid value for env is dev, test, sbx, or prod."
+    condition     = contains(["dev", "test", "sbx", "sandbox", "prod"], var.env)
```
carlpartridge, line 14: These are serving both legacy and gf right?  Do we need to do a ternary check (thinking of sbx vs sandbox)?

gsf: Adding "sandbox" while keeping "sbx" means this terraform will support both legacy and greenfield for the time being.

PR #263, opened 2025-07-15
PLT-1194: Create terraform service for shared KMS keys

christopher-maboh:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1194

## 🛠 Changes

Terraform service for shared KMS keys was created

## ℹ️ Context

We will save money and simplify configuration if we have a pair of standard KMS keys for each application environment.

### DIFF
--- a/terraform/services/kms-keys/main.tf
+++ b/terraform/services/kms-keys/main.tf
@@ -1,5 +1,5 @@
 module "standards" {
-  source      = "../../modules/standards"
+  source      = "github.com/CMSgov/cdap//terraform/modules/standards?ref=0bd3eeae6b03cc8883b7dbdee5f04deb33468260"
   app         = var.app
   env         = var.env
   root_module = "https://github.com/CMSgov/cdap/tree/main/terraform/services/kms-keys"


gsf (commented):
I noticed the workflow is "tf-kms-keys.yml". Please rename the service to match, placing files in terraform/services/tf-kms-keys and updating all references.

christopher-maboh:
> I noticed the workflow is "tf-kms-keys.yml". Please rename the service to match, placing files in terraform/services/tf-kms-keys and updating all references.

suggested changes made

terraform/services/kms-keys/terraform.tf
```diff
@@ -0,0 +1,26 @@
...
+  backend "s3" {
+    key = "kms-keys/terraform.tfstate"
+  }
+
+  required_providers {
+    aws = {
+      source                = "hashicorp/aws"
+      configuration_aliases = [aws.secondary]
+    }
+  }
```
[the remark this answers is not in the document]
mjburling: If a provider for a secondary region is necessary, I believe this is, too.

terraform/services/kms-keys/README.md
mjburling, line 1: With `terraform-docs` installed (e.g. `brew install terraform-docs`), you should be able to run `terraform-docs --config "$(git rev-parse --show-toplevel)/.terraform-docs.yml" .` from the module root to augment your README.md with some common documentation on dependencies, resources, variables, etc.

terraform/services/kms-keys/main.tf
```diff
@@ -0,0 +1,70 @@
+module "standards" {
+  source      = "../../modules/standards"
```
mjburling, line 2: @gsf I suspect there will be changes to the standards module and anywhere this module is referenced by local, relative path could make adopting changes in the future riskier or less appealing. Maybe we go ahead with a git reference on sha starting here?

gsf: Great idea. @christopher-maboh this should work:


```suggestion
  source      = "github.com/CMSgov/cdap//modules/standards?ref=0bd3eeae6b03cc8883b7dbdee5f04deb33468260"
```

mjburling: Try:
```hcl
  source      = "github.com/CMSgov/cdap//terraform/modules/standards?ref=0bd3eeae6b03cc8883b7dbdee5f04deb33468260"
```

mjburling (approved):

PR #267, opened 2025-07-22
PLT-1208: Aurora module for CDAP Customers

mjburling:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1208

## 🛠 Changes
Introduces a semi-opinionated Aurora module for CDAP customer usage.

## ℹ️ Context
Wider Aurora adoption and post-greenfield migration tasks.

## 🧪 Validation
- Applied and in-use across `ab2d` environments, including production

### DIFF
--- a/terraform/modules/aurora/README.md
+++ b/terraform/modules/aurora/README.md
@@ -130,12 +130,12 @@ No requirements.
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | <a name="input_backup_window"></a> [backup\_window](#input\_backup\_window) | Daily time range during which automated backups are created if automated backups are enabled in UTC, e.g. `04:00-09:00` | `string` | n/a | yes |
-| <a name="input_instance_class"></a> [instance\_class](#input\_instance\_class) | Aurora cluster instance class | `string` | n/a | yes |
+| <a name="input_instance_class"></a> [instance\_class](#input\_instance\_class) | Aurora cluster instance class, restricted to RI instances | `string` | n/a | yes |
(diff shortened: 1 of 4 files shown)


terraform/modules/aurora/README.md
```diff
@@ -0,0 +1,191 @@
+# Aurora Child Module
+This module creates the minimum set of semi-opinionated resources directly supporting creation and ongoing maintenance of an [Amazon Aurora DB Cluster](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/CHAP_Aurora.html) with ephemeral environment support.
+For a complete list of resources managed by this module, please see the [Resources section](#user-content-resources) below.
+
+## Important Usage Notes
+
+### Platform Module Expectations
+While emerging standardized `platform` isn't strictly required, usage is strongly recommended. **this** module expects the input variable `var.platform` as an object that includes the following, prescriptive fields:
+- `var.platform.is_ephemeral_env`
```
juliareynolds-nava, line 9: Should we prevent is_ephemeral if the env is prod?

mjburling: The boolean `is_ephemeral_env` comes from [the platform module's outputs](https://github.com/CMSgov/cdap/blob/8df211c48bc956bf0d1cb54fba68f67b751b0f6d/terraform/modules/platform/outputs.tf#L37-L41), and it's determined by the environment's name. If the names of the environment and the parent environment are equal, then the environment must be an enduring, path-to-production, non-ephemeral environment. If they're different, then the environment must be ephemeral.

```hcl
output "is_ephemeral_env" {
  description = "Returns true when environment is _ephemeral_, false when _established_"
  sensitive   = false
  value       = local.env != local.parent_env
}
```

And an ephemeral environment's parent environment (may be based off of) the `prod` environment. While they are temporary or _ephemeral_, they are still treated like any prod environment from access and security perspective, observing the same controls and so on, but they tend to use dev-like resources (fewer, smaller, less expensive) than those used in the `prod` environment.

PR #275, opened 2025-07-30
Jd/dpc 4821 opt out use only cluster

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4821

## 🛠 Changes

- Use aws_rds_cluster instead of aws_db_instance to find endpoint for DB_HOST environment variable for opt out lambdas
- Remove ab2d from opt out lambdas as no longer use
- Workflows to update lambdas only apply to app:env combination that have migrated to aurora

## ℹ️ Context

We are incrementally moving our databases to aurora clusters, which require a slightly different terraform resource reference. As the lambda infrastructure is stable, we are making the change in the existing services and blocking deploy except to aurora environments.
App-specific github workflows to allow bcda to use at their own pace when they are ready.

## 🧪 Validation

Manually ran in test environment, and testing the lambdas worked.


.github/workflows/tf-opt-out-import.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: ./actions/setup-tfenv-terraform
       - run: terraform fmt -check -diff -recursive terraform/services/opt-out-import
 
-  plan-apply:
+  plan-apply-dpc:
```
gsf, line 29:
```suggestion
  plan-apply:
```

gsf (approved):
Looks good! Thanks for removing ab2d-specific params to clean things up.

terraform/services/opt-out-import/main.tf
```diff
@@ -32,16 +17,15 @@ data "aws_iam_policy_document" "assume_bucket_role" {
   }
 }
 
-data "aws_db_instance" "this" {
-  db_instance_identifier = "${var.app}-${var.env}"
+data "aws_rds_cluster" "this" {
+  cluster_identifier = "${var.app}-${var.env}-aurora"
```
carlpartridge, line 21: Feels like this is the main change?  Is the issue that DPC is moving towards aurora, so we need to update this, but bcda/ab2d are not yet on aurora?  BCDA dev/test env should now be on Aurora (@jscott-nava confirm?), so can/should we expand the workflow matrix to inlclude BCDA?

terraform/services/opt-out-import/main.tf
```diff
@@ -32,16 +17,15 @@ data "aws_iam_policy_document" "assume_bucket_role" {
   }
 }
 
-data "aws_db_instance" "this" {
-  db_instance_identifier = "${var.app}-${var.env}"
+data "aws_rds_cluster" "this" {
```
carlpartridge, line 20: Nit: Is this still RDS or would a more generic identifier be better? (aws_db_cluster?)

terraform/services/opt-out-import/main.tf
gsf: Good point, we should be able to keep bcda dev and test in the matrix and check the plan output.

jdettmannnava: Added

carlpartridge (approved):

PR #277, opened 2025-08-01
[PLT-1250] remove ab2d-lambdas repo as it has been archived

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1250

## 🛠 Changes

Changed the code location of ab2d lambdas so that the github action role will have correct repo in the trust policy.

## ℹ️ Context

These changes were already made manually to fix errors described in PLT-1250, lambda github workflow with incorrect permissions.  This PR is to sync the terraform with the same repo change.

## 🧪 Validation

Tested via manual changes in non-prod.

### DIFF
--- a/terraform/modules/function/main.tf
+++ b/terraform/modules/function/main.tf
@@ -2,7 +2,7 @@ locals {
   provider_domain = "token.actions.githubusercontent.com"
   repos = {
     ab2d = [
-      "repo:CMSgov/ab2d-lambdas:*",
+      "repo:CMSgov/ab2d:*",
     ]
     bcda = [
       "repo:CMSgov/bcda-app:*",
--- a/terraform/services/github-actions-role/main.tf
+++ b/terraform/services/github-actions-role/main.tf
@@ -5,7 +5,6 @@ locals {
       "repo:CMSgov/cdap:*",
       "repo:CMSgov/ab2d-contracts:*",
       "repo:CMSgov/ab2d-events:*",
-      "repo:CMSgov/ab2d-lambdas:*",
       "repo:CMSgov/ab2d-properties:*",
       "repo:CMSgov/ab2d-website:*",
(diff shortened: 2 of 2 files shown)


gsf (dismissed):
Looks good! Reviewing the checks, I can see that the only tf plan where we see changes are ab2d environments in github-actions-role, like in this run for ab2d-prod: https://github.com/CMSgov/cdap/actions/runs/16679175367/job/47213416718?pr=277#step:6:54

terraform/modules/function/main.tf
```diff
@@ -2,7 +2,7 @@ locals {
   provider_domain = "token.actions.githubusercontent.com"
   repos = {
     ab2d = [
-      "repo:CMSgov/ab2d-lambdas:*",
+      "repo:CMSgov/ab2d:*",
```
mjburling, line 5: I may not have thought through this all the way, but is there anything ab2d-related that actually makes use of this module? If not, both the entry for `local.repos.ab2d` and references to "ab2d" in `var.app` could be eliminated entirely from the `function` module.

mjburling: > I'd prefer to leave it in as we are working to make this function module something the teams can use in their own ops code.

I don't disagree with a direction towards more re-use in modules, but I would argue the way in which this module is implemented, it may be better suited for sunsetting rather than iteration and continued development. The patterns in this module are a little marked, e.g. an initial dummy function defined that invites user to implement extra-terraform/extra-tofu automation for delivery/publication and deployment; function-specific KMS CMKs; reliance on hard-coded SSM values; difficult to image a transition to ephemeral environment support for this module.

In addition to reusability, I think we should be endeavoring to make the modules more composable to allow users to combine modules for their specific solutions where that makes sense... this module would need to be rewritten to accommodate composition.

mjburling (approved):

PR #278, opened 2025-08-06
PLT-1226: Extend alarm-to-slack service to cover all applications

christopher-maboh:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1226

## 🛠 Changes

alarm-to-slack service was extended to cover all applications

## ℹ️ Context

the is the need for  one instance of the alarm-to-slack function in each account should be able to handle routing of alarms to Slack for all applications and environments.

## 🧪 Validation

see checks


terraform/services/alarm-to-slack/lambda_src/lambda_function.py
```diff
@@ -34,65 +85,86 @@ def cloudwatch_message(record):
...
+            else:
+                log({'msg': f'AlarmName "{alarm_name}" does not match expected format',
+                     'messageId': record.get('messageId')})
                 return None
-            message['Emoji'] = ':checked:'
         else:
-            message['Emoji'] = ':anger:'
+            log({'msg': 'AlarmName not found in message',
+                 'messageId': record.get('messageId')})
+            return None
```
gsf, line 135: Please add tests that check the "AlarmName does not match" and "AlarmName not found" paths.

.github/workflows/tf-alarm-to-slack.yml
```diff
@@ -38,25 +38,44 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        app: [dpc]
-        env: [dev, test, sandbox, prod]
+        app: [bcda]
+        env: [dev, prod]
```
gsf, line 42: Envs should be test and prod

terraform/services/alarm-to-slack/main.tf
```diff
@@ -41,5 +37,12 @@ module "sns_to_slack_queue" {
...
-  sns_topic_arn = data.aws_sns_topic.cloudwatch.arn
+}
+
+resource "aws_sns_topic_subscription" "cloudwatch_alarms_to_queue" {
+  for_each   = data.aws_sns_topic.cloudwatch_alarms
+  topic_arn  = each.value.arn
+  protocol   = "sqs"
+  endpoint   = module.sns_to_slack_queue.arn
+  depends_on = [module.sns_to_slack_queue]
 }
```
gsf, line 48: Instead of creating topic subscriptions in this service, let the teams create these subscriptions in their terraform. Then they have more control over which topics the queue is subscribed to, and you don't have to pass in `app_envs` in this service.

terraform/services/alarm-to-slack/lambda_src/test_lambda_function.py
```diff
@@ -30,124 +31,170 @@ def test_cloudwatch_message_sqs_record():
...
+    }
     message = lambda_function.enriched_cloudwatch_message({
         'messageId': 'OK Sent',
-        'body': json.dumps({'Message': json.dumps(cloudwatch_message)})})
+        'body': json.dumps({'Message': json.dumps(cloudwatch_message)})
+    })
     assert message == enriched_cloudwatch_message
 
-@patch.dict(os.environ, {'IGNORE_OK': 'true'}, clear=True)
+@patch.dict(os.environ, {'IGNORE_OK_APPS': 'bcda'}, clear=True)
```
jdettmannnava, line 124: probably want 'bcda,dpc' to verify comma split

terraform/services/alarm-to-slack/main.tf
jdettmannnava: We already have control over this by which alarms we chose to send to sns, so I would rather be lazy and let you do this, fwiw

gsf: It's easier to create and manage these subscriptions in terraform applied in each environment.

terraform/services/alarm-to-slack/lambda_src/test_lambda_function.py
jdettmannnava: Well, we had some dissension about whether we wanted to get "all clear" messages; so although I myself prefer not to receive them, I wanted it to be easy for people to change their minds, especially on other teams.

terraform/services/alarm-to-slack/main.tf
jdettmannnava: Is this going to be one of those imported modules things?

terraform/services/alarm-to-slack/lambda_src/test_lambda_function.py
gsf: Yeah that sounds like a lot of noise in the slack alert channels. Let's hardcode the ignore until a team requires that functionality.

terraform/services/alarm-to-slack/lambda_src/lambda_function.py
```diff
@@ -5,26 +5,77 @@
...
+
+    try:
+        response = ssm_client.get_parameter(
+            Name=name,
+            WithDecryption=True
+        )
+        value = response['Parameter']['Value']
+        ssm_parameter_cache[name] = value
+        return value
+    except ClientError as e:
```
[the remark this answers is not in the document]
jdettmannnava: Yes, but the multiple returns is something I am a little iffy about. I would rather have:
if name not in ssm_parameter_cache:
...
return ssm_paramater_cache[name]

gsf (approved):
The function and queue modules need to be updated to avoid creating s3 buckets and KMS keys but that can be fixed in future work.

PR #279, opened 2025-08-07
PLT-1236: Add overrides for security groups and subnet groups, enable IAM auth by default, disable final snapshots, and copy tags to snapshots

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1236

## 🛠 Changes

- Adds variables to handle overrides on the default naming for security and subnet groups.
- Enables IAM auth by default.
- Enables copying tags from cluster instances to snapshots by default.
- Disables final snapshots.

## ℹ️ Context

As part of the DPC migration from RDS to Aurora clusters, we needed to override the naming enforced by the aurora module by default on some resources. We also have a security ask to enable IAM authentication, so this will enable that by default moving forward.

## 🧪 Validation

* dpc-test Aurora instance was deployed with this version of the module
* this is functional across ab2d `dev`, `test`, `sandbox`, and `prod`

### DIFF
--- a/terraform/modules/aurora/README.md
+++ b/terraform/modules/aurora/README.md
@@ -116,6 +116,17 @@ resource "aws_ssm_parameter" "writer_endpoint" {
      Manually updating sections between TF_DOCS tags may be overwritten.
      See https://terraform-docs.io/user-guide/configuration/ for more information.
 -->
+## Providers
+
+| Name | Version |
+|------|---------|
+| <a name="provider_aws"></a> [aws](#provider\_aws) | n/a |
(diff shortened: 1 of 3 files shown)


mjburling (approved):
These are some reasonable additions to make adoption a little easier. The changes here seem to work for `ab2d` just fine. Looks like this addresses `aws-foundational-security-best-practices-1.0.0/RDS.16` as expected, too.

Thanks, @gfreeman-navapbc!

gsf (approved):

PR #281, opened 2025-08-11
Jjr plt 1217 aws backups

juliareynolds-nava:
## 🎫 Ticket


terraform/services/backup-plan/primary.tf
```diff
@@ -0,0 +1,211 @@
...
+    completion_window = 180
+
+    lifecycle {
+      delete_after = 90
+    }
+  }
+
+  depends_on = [
+    aws_backup_vault.secondary_backup_vault
+  ]
```
gsf, line 182: This depends_on should not be necessary since the resource is referenced in the copy_action block above.

terraform/services/backup-plan/primary.tf
```diff
@@ -0,0 +1,211 @@
...
+
+  tags = {
+    cms-cloud-service = "AWS Backup"
+    Backup_Schedule   = "4hr1_d7_w35_m90"
+  }
```
gsf, line 187: These tags are not necessary.

terraform/services/backup-plan/primary.tf
```diff
@@ -0,0 +1,211 @@
...
+}
+
+resource "aws_backup_vault_policy" "primary_backup_vault_policy" {
+  for_each          = toset(local.apps)
+  backup_vault_name = aws_backup_vault.primary_backup_vault[each.value].name
+  policy            = data.aws_iam_policy_document.primary_backup_policy[each.value].json
+}
+
+resource "aws_backup_plan" "aws_backup_plan" {
```
gsf, line 125:
```suggestion
resource "aws_backup_plan" "this" {
```

terraform/services/backup-plan/primary.tf
```diff
@@ -0,0 +1,211 @@
...
+    aws_backup_vault.secondary_backup_vault
+  ]
+
+  tags = {
+    cms-cloud-service = "AWS Backup"
+    Backup_Schedule   = "4hr1_d7_w35_m90"
+  }
+}
+
+resource "aws_backup_selection" "aws_backup_selection" {
```
gsf, line 190:
```suggestion
resource "aws_backup_selection" "this" {
```

terraform/services/backup-plan/primary.tf
```diff
@@ -0,0 +1,211 @@
...
+  name         = "cdap_managed_backup_selection_${each.value}"
+  plan_id      = aws_backup_plan.aws_backup_plan[each.value].id
+  resources = [lower("arn:aws:rds:us-east-1:${data.aws_caller_identity.current.account_id}:cluster:${each.value}-${var.env}")]
+
+  selection_tag {
+    type  = "STRINGEQUALS"
+    key   = "Backup"
+    value = "cdap-managed"
+  }
+}
```
gsf, line 203: The backup selection must select backups by tag as required by CMS Cloud governance (see https://cloud.cms.gov/creating-custom-aws-backup-plans).

The resources attribute must be removed, and the selection_tag block must set the key as `AWS_Backup` and the value as `4Hours1CA_Daily7_Weekly35_Monthly90`.

gsf:
Instead of posting the terraform plan output in the PR description, create a tf-backup-plan workflow and point to checks.

terraform/services/backup-plan/primary.tf
```diff
@@ -0,0 +1,211 @@
...
+
+resource "aws_backup_vault_lock_configuration" "primary_vault_lock" {
+  for_each          = toset(local.apps)
+  backup_vault_name = aws_backup_vault.primary_backup_vault[each.value].name
+  changeable_for_days = 3
```
gsf, line 208: Drop this line to avoid creating a lock in compliance mode.

gsf:
Okay I've tested and if we continue to use the existing CMS_OIT_Backups_Vault in each account then all backups use our $app-$env KMS keys **except** for the cross-region snapshot in us-west-2, which is encrypted with the cms-oit-aws-backup-key KMS key. Furthermore, it appears we can select the $app-$env key for encryption of the database when we restore from that snapshot.

@mjburling Do you see this as reason enough to create our own backup vaults, or are there other reasons I'm missing for doing so?

mjburling:
> @mjburling Do you see this as reason enough to create our own backup vaults, or are there other reasons I'm missing for doing so?

Managing these resources shouldn't be too burdensome, but this seems like a decision you and your team can make. It's appealing to manage fewer resource, but it's also appealing to remember and/or document fewer exceptions.

terraform/services/backup-plan/terraform.tf
```diff
@@ -0,0 +1,15 @@
+terraform {
+  backend "s3" {
+    key = "backup-plan/terraform.tfstate"
+  }
+}
+
+provider "aws" {
+  alias  = "primary"
```
mjburling, line 8: I'm curious how it works when aliasing multiple providers, but only using one of the aliases. Does terraform just make the first defined provider the default? In any event, dropping the alias here is clearer, unless you wanted to explicitly define every _primary_ resource with the `aws.primary` provider.
```suggestion
```

mjburling (commented):
It's shaping up!

Though I'd prefer the KMS policies to be homed with the other policies in `kms-keys`, I'm OK with keeping some specific KMS key policies here, too.

In addition to some minor in-line adjustments that I've noted, some food for thought as I wonder about the vault names. `CMS-CDAP-MANAGED_VAULT_DPC` and `CMS-CDAP-MANAGED_VAULT_DPC_cr` look marked–is this format somehow required, or are we free to choose?
- when considering a list of AWS Backup Vaults, I don't think "CMS-CDAP-MANAGED_VAULT" helps us differentiate one resource from another
- I'm not sure if an input variable for the vault name is necessary–do you suspect we'll need to adjust vault names on the fly?
- a mix of underscores/underbars and strokes/dashes might be surprising to some (it surprised me)
- I'd propose using just the env and the app (`lower(each.value)`) to differentiate these resources, e.g. `dpc-prod` for DPC's production vault
- the alternative/secondary/DR region homed AWS Backup Vault with "_cr" seems unnecessary–I think it can safely share a name with its primary counterpart

gsf:
> Managing these resources shouldn't be too burdensome, but this seems like a decision you and your team can make. It's appealing to manage fewer resource, but it's also appealing to remember and/or document fewer exceptions.

I think documentation of the need to restore the database under the shared key is worth doing in any case.

@juliareynolds-nava Let's drop the vaults specific to application and environment and use the existing CMS_OIT_Backups_Vault in each account.

PR #282, opened 2025-08-12
PLT-1228: SecHub warning: This control fails if kms:Decrypt or kms:ReEncryptFrom actions are allowed on all KMS keys in an inline policy.

christopher-maboh:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1228

## 🛠 Changes

inline policy for import and export role was changed. Restricting the kms key action to the specific kms key being used.

## ℹ️ Context

we received a SecHub warning on allowing the AWS Key Management Service (KMS) decryption actions on all KMS keys. This control fails if kms:Decrypt or kms:ReEncryptFrom actions are allowed on all KMS keys in an inline policy.

## 🧪 Validation

See checks

### DIFF
--- a/terraform/modules/function/main.tf
+++ b/terraform/modules/function/main.tf
@@ -83,7 +83,7 @@ resource "aws_kms_alias" "env_vars" {
 
 data "aws_caller_identity" "current" {}
 
-data "aws_iam_policy_document" "function_policy_document" {
+data "aws_iam_policy_document" "default_function" {
   statement {
     actions = [
       "ec2:CreateNetworkInterface",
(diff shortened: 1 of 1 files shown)


juliareynolds-nava (dismissed):

juliareynolds-nava (dismissed):

gsf (approved):

gsf:
Looks like dpc-test lambdas will need to be updated for apply to succeed. Check with @gfreeman-navapbc on appropriate references.

gfreeman-navapbc:
The DPC Aurora clusters will have the identifier of `dpc-<env>`, currently just in the test environment, but should also be in dev soon

gsf (approved):
Looks good! Make sure to check the apply since the switch from `inline_policy` to `aws_iam_role_policy` might encounter some aws api weirdness.

PR #299, opened 2025-09-05
PLT-1282: add cdap app and env in bucket module

christopher-maboh:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1282

## 🛠 Changes
workflow terraform apply failed with is error message
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Invalid value for variable
│ 
│   on main.tf line 9, in module "tfstate_bucket":
│    9:   app = var.app
│     ├────────────────
│     │ var.app is "cdap"
│ 
│ Valid value for app is ab2d, bcda, or dpc.
│ 
│ This was checked by the validation rule at
│ ../../modules/bucket/variables.tf:4,3-13.
╵
╷
│ Error: Invalid value for variable
│ 
│   on main.tf line 10, in module "tfstate_bucket":
│   10:   env = var.env
│     ├────────────────
│     │ var.env is "mgmt"
│ 
│ Valid value for env is dev, test,sandbox, or prod.
│ 
│ This was checked by the validation rule at
│ ../../modules/bucket/variables.tf:13,3-13.
╵

Error: Process completed with exit code 1.
CDAP env and app added to bucket module variables.tf

## ℹ️ Context

w


terraform/modules/bucket/variables.tf
```diff
@@ -1,18 +1,18 @@
...
 
 variable "env" {
-  description = "The application environment (dev, test, sandbox, prod)"
+  description = "The application environment (dev, test, sbx, sandbox, prod, mgmt)"
   type        = string
   validation {
-    condition     = contains(["dev", "test", "sandbox", "prod"], var.env)
-    error_message = "Valid value for env is dev, test,sandbox, or prod."
+    condition     = contains(["dev", "test", "sbx", "sandbox", "prod", "mgmt"], var.env)
+    error_message = "Valid value for env is dev, test, sbx, sandbox, prod, or mgmt."
```
gfreeman-navapbc, line 15: Why are we adding `sbx` back here? AFAIK we are now standardized on `sandbox` across the board

christopher-maboh: > Why are we adding `sbx` back here? AFAIK we are now standardized on `sandbox` across the board

yeah let me take that out. I will need to take it out from the tfstate service as well. I copied this variables.tf file from there since the error was pointing to it and I didn't realize it still had sbx on it

gfreeman-navapbc (approved):

PR #301, opened 2025-09-09
[PLT-1217] AWS Backups

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1217

## 🛠 Changes

added aws backup plan, vaults, policies

When applying we must import existing vaults:
terraform import aws_backup_vault.primary CMS_OIT_Backups_Vault 
terraform import aws_backup_vault.secondary CMS_OIT_Backups_Vault

## ℹ️ Context

Cleaned up PR from #281.

- [x] 4hr1CA_d7_w35_m90 plan, adding cross-region to the 4hourly_1 rule. Result: 6 4-hour backups, 7 daily, 5 weekly, and 3 monthly.
- [x] Create a service in the cdap repo to manage that backup plan in each account.
- [x] Update terraform for each team to use the new backup plan and drop any other snapshots from the database configuration.  Current selection is an example that selects clusters by app-env name AND for tag:  Backup = cdap-managed.  **We should refine how we plan to select resources to be backed up.**
- [x] Per Michael Burling, create a vault for each ADO in both test and prod. 
- [x]  Per Sean Fern, Create mirror vaults for each ADO in us-west-2.  The mirror vault names in secondary have a suffix "_cr".  This is for uniqueness so that vault locks apply correctly.
- [x]  Per Michael Burling, Each vault should use the same key as the database being backed up.
- [x]  Per Michael Burling, Add a lock to each Vault configured like the CMS_OIT vault.
- [x] Per Sean Fern, we are using the role:  arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/delegatedadmin/developer/**cms-oit-aws-backup-service-role**

## 🧪 Validation

See checks - tf-backup-plan workflow.

### DIFF
--- a/.github/workflows/tf-backup-plan.yml
+++ b/.github/workflows/tf-backup-plan.yml
@@ -1,0 +1,51 @@
+name: tf-backup-plan
+run-name: tf-backup-plan ${{ (inputs.apply || (github.event_name == 'push' && github.ref == 'refs/heads/main')) && 'apply' || 'plan' }}
+
+on:
+  push:
+    paths:
+      - .github/workflows/tf-backup-plan.yml
+      - terraform/services/backup-plan/**
+  workflow_dispatch:
+    inputs:
+      apply:
+        required: false
+        type: boolean
+        description: "Apply the terraform?"
+
+jobs:
+  check-fmt:
+    runs-on: codebuild-cdap-${{github.run_id}}-${{github.run_attempt}}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./actions/setup-tfenv-terraform
+      - run: terraform fmt -check -diff -recursive terraform/services/backup-plan
+
+  plan-apply:
+    needs: check-fmt
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: codebuild-cdap-${{github.run_id}}-${{github.run_attempt}}
+    defaults:
+      run:
+        working-directory: ./terraform/services/backup-plan
+    strategy:
+      fail-fast: false
+      matrix:
+        app: [bcda]
+        env: [test, prod]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./actions/setup-tfenv-terraform
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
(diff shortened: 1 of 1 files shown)


terraform/services/backup-plan/main.tf
```diff
@@ -0,0 +1,79 @@
...
+
+
+data "aws_backup_vault" "secondary" {
+  name     = "CMS_OIT_Backups_Vault"
+  provider = aws.secondary
+}
+
+resource "aws_backup_plan" "this" {
+  name = "4Hours1CA_Daily7_Weekly35_Monthly90"
+  #only the 4hr rule should be copied to secondary
```
gsf, line 21:
```suggestion

  # The 4 hour rule copies to the west region for disaster recovery
```

terraform/services/backup-plan/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  rule {
+    rule_name         = "4Hours1CA"
```
gsf, line 23: I've just realized the `CA` I suggested for this rule is not appropriate as this does not copy "Cross Account" but cross-region. A more fitting name would be `4Hours1DR` to denote "Disaster Recovery". This should be changed here, in the backup plan name, and in the selection tag.

terraform/services/backup-plan/main.tf
```diff
@@ -0,0 +1,78 @@
...
+resource "aws_backup_selection" "this" {
+  # This iam role is the one CMS is using for their backup plan.
+  iam_role_arn = "arn:aws:iam::${module.standards.account_id}:role/delegatedadmin/developer/cms-oit-aws-backup-service-role"
+  name         = "cdap_managed_backup_selection"
+  plan_id      = aws_backup_plan.this.id
+
+  selection_tag {
+    type  = "STRINGEQUALS"
+    key   = "AWS_Backup"
+    value = "4Hours1DR_Daily7_Weekly35_Monthly90"
```
gsf, line 76: I had also missed that the selection tags on OIT backup plans use a shortened value here. We should match that for consistency.


```suggestion
    value = "4hr1dr_d7_w35_m90"
```

gsf (dismissed):

gsf (approved):

PR #302, opened 2025-09-09
PLT-1279 : Move CDAP over to OpenTofu

christopher-maboh:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1279

## 🛠 Changes

updates where made to workflows to use openTofu

## ℹ️ Context

The is the need to Move CDAP over to OpenTofu

## 🧪 Validation

see checks
last successful runs for all workflows https://github.com/CMSgov/cdap/actions

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -10,4 +10,4 @@ In **services** we have terraform split into functional units. For background on
 
 ## Version
 
-Note that the tofu version is set in [.opentofu-version](.opentofu-version). If you use [tfenv](https://github.com/tfutils/tfenv) then the correct version of opentofu will be installed and used automatically.
+Note that the tofu version is set in [.opentofu-version](.opentofu-version). If you use [tenv](https://github.com/tofuutils/tenv) then the correct version of opentofu will be installed and used automatically.


.github/workflows/tf-admin-aco-deny.yml
```diff
@@ -18,39 +18,46 @@ on:
...
   check-fmt:
     runs-on: codebuild-cdap-${{github.run_id}}-${{github.run_attempt}}
     steps:
       - uses: actions/checkout@v4
-      - uses: ./actions/setup-tfenv-terraform
-      - run: terraform fmt -check -diff -recursive terraform/services/admin-aco-deny
+      - uses: sigstore/cosign-installer@d58896d6a1865668819e1d91763c7751a165e159 # v3.9.2
+      - uses: ./actions/setup-tenv # TODO Use cmsgov/cdap/actions/setup-tenv@<hash> after this lands in main
+        env:
+          TENV_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
gsf, line 33: Good attempt to fix the rate-limiting errors, but they're occuring on the tofu run steps. I think the best way to solve this will be to set the TENV_GITHUB_TOKEN env var at the top level of the workflow, above the defaults block above.

terraform/README.md
```diff
@@ -10,4 +10,4 @@ In **services** we have terraform split into functional units. For background on
 
 ## Version
 
-Note that the terraform version is set in [.terraform-version](.terraform-version). If you use [tfenv](https://github.com/tfutils/tfenv) then the correct version of terraform will be installed and used automatically.
+Note that the tofu version is set in [.opentofu-version](.opentofu-version). If you use [tfenv](https://github.com/tfutils/tfenv) then the correct version of opentofu will be installed and used automatically.
```
gsf, line 13:
```suggestion
Note that the tofu version is set in [.opentofu-version](.opentofu-version). If you use [tenv](https://github.com/tofuutils/tenv) then the correct version of opentofu will be installed and used automatically.
```

gsf (approved):

PR #305, opened 2025-09-10
[PLT-1299] Fargate ecs service and task

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1299

## 🛠 Changes

Added terraform module for creating Fargate ecs service and task.

## ℹ️ Context

To support migration from ec2 to Fargate.

## 🧪 Validation

See Validation section at https://github.com/CMSgov/ab2d/pull/1623.


terraform/modules/service/outputs.tf
```diff
@@ -0,0 +1,10 @@
+output "aws_ecs_service" {
+  description = "The ecs service for the given inputs."
+  value       = aws_ecs_service.this
+}
+
+output "aws_ecs_task_definition" {
+  description = "The ecs task definition for the given inputs."
+  value       = aws_ecs_task_definition.this
+}
+
```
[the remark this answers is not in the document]
gsf: This comment was in reference to the output names, so "service" instead of "aws_ecs_service" and "task_definition" instead of "aws_ecs_task_definition".

terraform/modules/service/variables.tf
```diff
@@ -0,0 +1,156 @@
...
+
+variable "propagate_tags" {
+  default     = "SERVICE"
+  description = "Determines whether to propagate the tags from the task definition to the Amazon EBS volume."
+  type        = string
+  validation {
+    condition     = contains(["SERVICE", "TASK_DEFINITION"], var.propagate_tags)
+    error_message = "Invalid propagate_tags setting. Must be 'SERVICE' or 'TASK_DEFINITION'"
+  }
+}
```
gsf, line 113: This doesn't need to be a variable. Set it in the module.

juliareynolds-nava: defaulted to service

terraform/modules/service/variables.tf
```diff
@@ -0,0 +1,153 @@
...
+
+variable "task_execution_role_arn" {
+  default     = null
+  description = "ARN of the task execution role that the Amazon ECS container agent and the Docker daemon can assume.  Defaults to creation of a new role."
+  type        = string
+}
+
+variable "task_app_role_arn" {
+  description = "ARN of IAM role that allows your Amazon ECS container task to make calls to other AWS services."
+}
```
gsf, line 135: Why is this `task_app_role_arn` instead of `task_role_arn`? Also, type should be set to string.

terraform/modules/service/variables.tf
```diff
@@ -0,0 +1,153 @@
...
+  description = "When *changed* to `true`, trigger a new deployment of the ECS Service even when a deployment wouldn't otherwise be triggered by other changes. **Note**: This has no effect when the value is `false`, changed to `false`, or set to `true` between consecutive applies."
+  type        = bool
+}
+
+
+variable "image" {
+  description = "The image used to start a container. This string is passed directly to the Docker daemon. By default, images in the Docker Hub registry are available. Other repositories are specified with either `repository-url/image:tag` or `repository-url/image@digest`"
+  type        = string
+  default     = null
+}
```
gsf, line 51: I think the image tag should be required. Is there a use case for having a service and task without an image?

terraform/modules/service/variables.tf
```diff
@@ -0,0 +1,153 @@
...
+    target_group_arn = string
+    container_name   = string
+    container_port   = number
+  }))
+}
+
+variable "memory" {
+  description = "Amount (in MiB) of memory used by the task."
+  type        = number
+}
```
gsf, line 65: Put memory next to cpu in this file and add a comment referencing https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size

terraform/modules/service/main.tf
```diff
@@ -0,0 +1,125 @@
...
+      },
+    ]
+  })
+}
+
+resource "aws_iam_role_policy" "ecs_task_execution_policy_attachment" {
+  name   = "ecs-task-execution-policy"
+  role   = aws_iam_role.ecs_task_execution_role.name
+  policy = data.aws_iam_policy_document.ecs_task_execution_policy.json
+}
```
gsf, line 125: The three resources above can be defined with less redundancy:

* aws_iam_policy_document.execution
* aws_iam_role.execution
* aws_iam_role_policy.execution

Also, the names on the role and policy must be unique to the task. I would suggest `"${aws_ecs_task_definition.this.family}-execution"` as the name for both.

Finally, these resources should only be created if the execution_role_arn variable is not set.

terraform/modules/service/main.tf
```diff
@@ -0,0 +1,125 @@
...
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:BatchGetImage",
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+      "ssm:GetParameters"
+    ]
+    resources = ["*"]
+  }
+}
```
gsf, line 103: I believe this policy will need kms:Decrypt on our shared KMS keys to allow access to ssm params. See https://github.com/CMSgov/dpc-ops/blob/25e1d52ea29fb0d9ecc06c448f4faa64c0094e2a/terraform/modules/resources/ecs-service/main.tf#L508-L512

terraform/modules/service/variables.tf
```diff
@@ -0,0 +1,153 @@
...
+  default = null
+}
+
+variable "service_name_override" {
+  default     = null
+  description = "Desired service name for the service tag on the aws ecs service.  Defaults to platform.service."
+  type        = string
+}
+
+variable "task_execution_role_arn" {
```
gsf, line 127:
```suggestion
variable "execution_role_arn" {
```

terraform/modules/service/variables.tf
```diff
@@ -0,0 +1,153 @@
...
+  type = list(object({
+    subnets          = list(string)
+    assign_public_ip = string
+    security_groups  = list(string)
+  }))
+}
+
+variable "platform" {
+  description = "Object that describes standardized platform values."
+  type        = any
```
gsf, line 105: Set type to object. See https://github.com/CMSgov/cdap/pull/268/files#diff-8e04d31622ba5ffa53a573b60074ebc6a35c310f36c4bd28c5f3857ad9da571fR32

gsf:
Please add validation to the PR description. See https://github.com/CMSgov/cdap/pull/268 for a good example for how to do so for a new module.

terraform/modules/service/main.tf
gsf: Where did you use the key from the platform modulet? The policy still needs the allow action I pointed to in the link.

terraform/modules/service/main.tf
```diff
@@ -0,0 +1,125 @@
+resource "aws_ecs_task_definition" "this" {
+  family             = var.family_name_override != null ? var.family_name_override : "${var.platform.env}-${var.platform.app}-${var.platform.service}"
```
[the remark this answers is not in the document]
gsf: Do you have changes to push up still? This comment does not appear to be addressed.

terraform/modules/service/main.tf
juliareynolds-nava: allow action is added.

terraform/modules/service/variables.tf
```diff
@@ -0,0 +1,132 @@
...
+
+variable "execution_role_arn" {
+  description = "ARN of the task execution role that the Amazon ECS container agent and the Docker daemon can assume.  Defaults to creation of a new role."
+  type        = string
+  default     = null
+}
+
+variable "family_name_override" {
+  default     = null
+  description = "The desired family name for the ECS task definition.  If null will default to service name."
```
gsf, line 49: Missing type here.

terraform/modules/service/README.md
```diff
@@ -0,0 +1,154 @@
...
+    hostPort           = "8080" # Optional
+    name               = "portName" # Optional
+    protocol           = "tcp" # Optional
+  }]
+  propagate_tags = "SERVICE"
+  # SERVICE: Tags defined on the aws_ecs_service resource itself will be propagated to the tasks. Default value.
+  # TASK_DEFINITION: Tags defined on the aws_ecs_task_definition resource will be propagated to the tasks.
+  service_name_override = "my_test_service" # Optional - Desired service name for the service tag on the aws ecs service.  Defaults to platform.service.
+  execution_role_arn = "this_is_an_iam_role_arn" #Optional - ARN of the task execution role that the Amazon ECS container agent and the Docker daemon can assume.  Defaults to creation of a new role.
+  app_role_arn = "this_is_an_iam_role_arn"  # ARN of IAM role that allows your Amazon ECS container task to make calls to other AWS services.
```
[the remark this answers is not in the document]
gsf: Good point. The differences between these roles is an ongoing source of confusion because it is tricky to clearly define them. Let me see if I can more clearly do so.

terraform/modules/service/README.md
```diff
@@ -0,0 +1,154 @@
...
+    [{
+      file_system_id = "fs-12345678"
+      root_directory = "/app_data"
+      authorization_config = [{
+        iam = "ENABLED"
+      }]
+    }]
+    }]
+}
+
```
gsf, line 88: This section needs to be updated. I'll do that.

gsf (approved):

michaeljvaldes (approved):

PR #307, opened 2025-09-15
BCDA-9400: add lifecycle transition tags to bucket objects

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9400

## 🛠 Changes

- Added lifecycle management to S3 buckets that have a tag of `lifecycle-transition` and value set to `ia`
- When new versions of objects are uploaded (like new Lambda deployment packages), the previous versions will automatically transition to Standard-IA storage after 3 days

## ℹ️ Context

We were not using lifecycle transitions on our buckets, and a Security Hub control failed. Remediating it by adding a basic lifecycle transition.

## 🧪 Validation

Using the AWS console when the Lambda is deployed into lower environments.

### DIFF
--- a/terraform/modules/bucket/main.tf
+++ b/terraform/modules/bucket/main.tf
@@ -130,7 +130,7 @@ resource "aws_s3_bucket_lifecycle_configuration" "this" {
     }
 
     noncurrent_version_transition {
-      noncurrent_days = 3
+      noncurrent_days = 30
       storage_class   = "STANDARD_IA"
     }
   }


jscott-nava (approved):

gsf:
> * Don't use the tag filter in TF/bucket
> * Add a variable to the bucket Terraform (for any callers to enable lifecycle transition if they would like to)
> * The function module can pass this variable value to the bucket module when calling it

Yes, that's correct! And apologies for the back and forth on this. Longer term, we should move the terraform for these functions into an ops directory in the bcda-app repo, and any changes we need to make to the bucket module will more clearly be the responsibility of CDAP.

PR #310, opened 2025-09-17
DPC-4813 change db port

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4813

## 🛠 Changes

- added port_override variable to aurora module
- updated api-rds, api-waf-sync, opt-out-export, and opt-out-import services to have security group allow port 5431 for dpc.

## ℹ️ Context

Sechub recommends not using default port for db (RDS.23). This ticket will allow changes in dpc-ops and dpc-app to support this change.

Keeping 5432 in the port range for the dpc lambda so current code does not fail.

## 🧪 Validation

All tests pass.


terraform/modules/aurora/variables.tf
```diff
@@ -141,3 +141,9 @@ variable "security_group_override" {
   description = "Override for the security group name"
   type        = string
 }
+
+variable "port_override" {
```
gfreeman-navapbc, line 145: Since we're adding a new variable, could you run `terraform-docs --config "$(git rev-parse --show-toplevel)/.terraform-docs.yml" .` to generate the new README for the aurora module?

juliareynolds-nava (approved):

jdettmannnava:
Closed as exemption makes unnecessary

PR #311, opened 2025-09-17
[PLT-1350] Removing legacy api-rds workflow and service.

jscott-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1350

## 🛠 Changes

This PR removes the TF code for the legacy api-rds service along with the GitHub Actions workflow that calls it.

## ℹ️ Context

With the move to Aurora complete the api-rds service and workflow are now obsolete and should be removed.

## 🧪 Validation

There is no good way to validate this change beforehand since upon merge it will simply remove the existing workflow and associated services code, but searches for 'api-rds' and 'rds' in the cdap/terraform directory of the CDAP repo revealed that no non-documentation references to the api-rds service or workflow remain.


mjburling (approved):
💯

juliareynolds-nava (approved):

PR #312, opened 2025-09-18
[PLT-1348] Remove lifecycle-transition filter

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1348

## 🛠 Changes

Removed lifecycle transition filter on bucket module lifecycle configuration.  This is to ensure that non-current objects in all buckets transition to IA after 30 days. 
## ℹ️ Context

This change is to conform with security hub findings regarding bucket lifecycle rules.

## 🧪 Validation

See checks.


gsf (commented):
This PR should also include reverting the change to workflows in https://github.com/CMSgov/cdap/pull/307

terraform/modules/bucket/main.tf
```diff
@@ -122,13 +122,6 @@ resource "aws_s3_bucket_lifecycle_configuration" "this" {
     id     = "noncurrent-ia-tagged"
```
gsf, line 122:
```suggestion
    id     = "noncurrent-ia"
```

PR #313, opened 2025-09-18
[PLT-1332] Adding cron trigger to service workflows.

jscott-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1332

## 🛠 Changes

This PR adds a cron trigger of '12 14 * * 1-5' to all TF service workflows to ensure that they are run at 9:12am or 10:12am (depending on daylight savings) every weekday.

## ℹ️ Context

This change is being made to ensure that Terraform drift in any CDAP service will be caught and corrected within one business day.

## 🧪 Validation

When these changes were pushed to a remote branch the 'on push' trigger ran all affected workflows successfully. While this did not exercise the cron triggers directly, it shows that the workflows remain syntactically correct with the addition of the cron triggers.

### DIFF
--- a/.github/workflows/tf-admin-aco-deny.yml
+++ b/.github/workflows/tf-admin-aco-deny.yml
@@ -13,7 +13,7 @@ on:
       - terraform/modules/vpc/**
       # - terraform/services/admin-aco-deny/**
   schedule:
-    - cron: "*/30 * * * *"
+    - cron: "*/5 * * * *"
   workflow_dispatch:
     inputs:
       apply:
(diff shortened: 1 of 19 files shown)


gsf (commented):
One thing I forgot to account for in the ticket description is that these jobs need to run the apply when triggered by the cron schedule. This unfortunately means that the conditional in the run-name and that last step will need to be updated for the cron trigger. We may be able to simplify it behind an env variable or something but I recall running into limits with run-name when trying to do that previously.

juliareynolds-nava (approved):

juliareynolds-nava (approved):

gsf (approved):

PR #314, opened 2025-09-23
[PLT-1227] Cost Anomaly alert to slack

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1227

## 🛠 Changes

Added a cost anomaly cost monitor and sns topic subscription.  This is the first part of required work for this feature.  A lambda to push alerts to slack will be completed in PLT-1361

## ℹ️ Context

Users want to be alerted in Slack about significant cost fluctuations, so that we can take actions on addressing potential cost anomalies.

## 🧪 Validation

See check here:  https://github.com/CMSgov/cdap/actions/runs/19306698285

### DIFF
--- a/terraform/services/cost-anomaly/tofu.tf
+++ b/terraform/services/cost-anomaly/tofu.tf
@@ -4,18 +4,6 @@ terraform {
   }
 }
 
-variable "region" {
-  default  = "us-east-1"
-  nullable = false
-  type     = string
-}
-
-variable "secondary_region" {
-  default  = "us-west-2"
-  nullable = false
-  type     = string
-}
-
 provider "aws" {
   region = "us-east-1"
   default_tags {


.github/workflows/tf-cost-anomaly.yml
```diff
@@ -0,0 +1,61 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: sigstore/cosign-installer@d58896d6a1865668819e1d91763c7751a165e159 # v3.9.2
+      - uses: cmsgov/cdap/actions/setup-tenv@8343fb96563ce4b74c4dececee9b268f42bd4a40
+      - uses: aws-actions/configure-aws-credentials@a03048d87541d1d9fcf2ecf528a4a65ba9bd7838 # v5.0.0
+        with:
+          role-to-assume: arn:aws:iam::${{ contains(fromJSON('["dev", "test"]'), matrix.env) && secrets.NON_PROD_ACCOUNT || secrets.PROD_ACCOUNT }}:role/delegatedadmin/developer/${{ matrix.app }}-${{ matrix.env }}-github-actions
+          aws-region: ${{ vars.AWS_REGION }}
+      - run: tofu init -backend-config=../../backends/${{ matrix.app }}-${{ matrix.env }}.s3.tfbackend
+      - run: tofu plan -exclude=module.sns_to_slack_queue.data.aws_iam_policy_document.sns_send_message -exclude=module.sns_to_slack_queue.aws_sns_topic_subscription.this -out=tf.plan
```
gsf, line 54: Why are these excludes in the plan?

terraform/services/cost-anomaly/variables.tf
```diff
@@ -0,0 +1,20 @@
...
+  description = "List of IAM policy documents that are merged together into the exported document. Statements must have unique `sid`s"
+  type        = list(string)
+  default     = []
+}
+
+variable "override_policy_documents" {
+  description = "List of IAM policy documents that are merged together into the exported document. In merging, statements with non-blank `sid`s will override statements with the same `sid`"
+  type        = list(string)
+  default     = []
+}
```
gsf: The source and override policy documents should not be variables in the service. They should be defined in the service and passed to the module.

terraform/services/cost-anomaly/terraform.tf
```diff
@@ -0,0 +1,27 @@
+terraform {
+  backend "s3" {
+    key = "cost-anomaly/terraform.tfstate"
+  }
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~>5"
+    }
+  }
```
gsf: Drop required_providers in the service

terraform/modules/queue/main.tf
```diff
@@ -47,6 +47,7 @@ data "aws_iam_policy_document" "sns_send_message" {
       values   = [var.sns_topic_arn]
     }
   }
+  source_policy_documents = var.policy_documents
```
gsf: The issue with just adding source_policy_documents here is that the policy document will only be created if the sns_topic_arn variable is set.

We could remove the count conditional (here and in aws_sqs_queue_policy below) and associate the policy document regardless of whether sns_topic_arn is set, but that might break existing usage of the queue module by creating and associating a policy where that hadn't been done before.

I believe it will be safest to expand the count conditionals in these resources to include the source_policy_documents variable.

terraform/modules/queue/main.tf
```diff
@@ -47,10 +47,11 @@ data "aws_iam_policy_document" "sns_send_message" {
       values   = [var.sns_topic_arn]
```
gsf: Let's remove the above statement and count conditionals (and remove the aws_sns_topic_subscription resource below as well.). Then update [cclf-import and opt-out-import](https://github.com/search?q=repo%3ACMSgov%2Fcdap+sns_topic_arn+path%3A%2F%5Eterraform%5C%2Fservices%5C%2F%2F&type=code) to pass in a policy document as you've done in cost-anomaly.

terraform/modules/queue/variables.tf
gsf: The sns_topic_arn variable can be dropped

terraform/services/cost-anomaly/tofu.tf
```diff
@@ -0,0 +1,33 @@
...
+}
+
+variable "secondary_region" {
+  default  = "us-west-2"
+  nullable = false
+  type     = string
+}
+
+provider "aws" {
+  region = var.region
```
gsf: Drop the variables above and just set the region in the provider.


```suggestion
  region = "us-east-1"
```

terraform/services/cost-anomaly/tofu.tf
```diff
@@ -0,0 +1,33 @@
...
+  default_tags {
+    tags = local.default_tags
+  }
+}
+
+provider "aws" {
+  alias = "secondary"
+
+  region = var.secondary_region
```
gsf:
```suggestion
  region = "us-west-2"
```

terraform/services/cost-anomaly/tofu.tf
```diff
@@ -0,0 +1,33 @@
...
+  default  = "us-east-1"
+  nullable = false
+  type     = string
+}
+
+variable "secondary_region" {
+  default  = "us-west-2"
+  nullable = false
+  type     = string
+}
```
gsf: Drop the variables here

terraform/services/cclf-import/main.tf
```diff
@@ -65,8 +65,34 @@ module "cclf_import_queue" {
...
+      values   = [data.aws_ssm_parameter.bfd_sns_topic_arn.value]
+    }
+  }
+}
+
+resource "aws_sns_topic_subscription" "this" {
+  endpoint  = module.cclf_import_queue.arn
+  protocol  = "sqs"
+  topic_arn = data.aws_ssm_parameter.bfd_sns_topic_arn.value
 }
```
gsf: I'm seeing here that these changes will result in destroying and remaking the policy and topic subscription: https://github.com/CMSgov/cdap/actions/runs/19314118921/job/55241472316?pr=314

It could be avoided by adding a `moved` block here but I think it's fine to cycle the resources. Nothing stateful in them and we should be able to create the subscription again as long as the queue ARN hasn't changed.

gsf (approved):

jscott-nava (approved):

PR #318, opened 2025-09-26
PLT-1342: Set default force_ssl cluster parameter in aurora module

christopher-maboh:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1342

## 🛠 Changes

default force_ssl cluster parameter has been added in aurora module

## ℹ️ Context

need to standardize default force_ssl cluster parameter across all apis, with shared aurora module in place. its best to have this in module and api teams consume module rather than have it set in apis infra directly.

## 🧪 Validation

The Terraform plan for the Aurora module changes has been validated in [BCDA-OPS PR #1268](https://github.com/CMSgov/bcda-ops/pull/1268).

### DIFF
--- a/terraform/modules/aurora/main.tf
+++ b/terraform/modules/aurora/main.tf
@@ -40,6 +40,12 @@ resource "aws_rds_cluster_parameter_group" "this" {
   family      = local.aurora_family
   description = "Aurora cluster parameter group for ${local.service_prefix}"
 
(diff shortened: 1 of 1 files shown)


gsf (commented):
Looks good. This should be validated with a plan in an application team's terraform. Create PRs in those repos and point to one of those.

PR #322, opened 2025-10-01
PLT-1224

mjburling:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1224

### DIFF
--- a/terraform/backends/cdap-insights-mgmt.s3.hcl
+++ b/terraform/backends/cdap-insights-mgmt.s3.hcl
@@ -1,0 +1,2 @@
+bucket       = "cdap-mgmt-tfstate-20250930180004007700000001"
+use_lockfile = true
--- a/terraform/modules/bucket/main.tf
+++ b/terraform/modules/bucket/main.tf
@@ -1,3 +1,9 @@
+locals {
+  cdap_ssm = zipmap(
+    data.aws_ssm_parameters_by_path.cdap.names,
+    data.aws_ssm_parameters_by_path.cdap.values
+  )
+  access_logs_bucket = lookup(local.cdap_ssm, "/cdap/bucket-access-logs-bucket", null)
 }
 
 resource "aws_s3_bucket" "this" {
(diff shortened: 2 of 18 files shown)


terraform/backends/cdap-insights-mgmt.s3.hcl
mjburling, line 1: Obviously, we don't use the `hcl` suffix on the other backends, so this is non-standard. However, this would be a nice touch to enable native highlighting/formatting in IDEs, etc.

terraform/modules/bucket/main.tf
```diff
@@ -99,19 +100,24 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "this" {
   }
 }
 
-data "aws_iam_account_alias" "current" {}
+data "aws_ssm_parameters_by_path" "cdap" {
+  path = "/cdap"
+  recursive = true
+}
```
mjburling, line 106: Using the `aws_ssm_parameters_by_path` data source allows for empty results. Paired with the call to the `zipmap()` in `local.cdap_ssm` **and** the `lookup()` on the same for `local.access_logs_bucket`, this helps us avoid failures on missing configuration and obviates the need for hard-coding bucket-access-logs buckets.

terraform/modules/standards/outputs.tf
```diff
@@ -10,12 +10,6 @@ output "service" {
   value       = local.service
 }
 
-output "region_name" {
-  description = "**Deprecated**. Use `primary_region.name`. The region name associated with the current caller identity"
-  sensitive   = false
-  value       = data.aws_region.this.name
-}
-
```
mjburling, line 18: As noted, this was already deprecated in favor of the `primary_region` output. With updates to aws provider version 6, `data.aws_region` has a deprecated `name` and an attribute, in favor of `region`, so... references like this would be most correct as `data.aws_region.this.region`. ¯\_(ツ)_/¯

terraform/modules/standards/outputs.tf
```diff
@@ -29,11 +23,17 @@ output "secondary_region" {
 }
 
 output "account_id" {
-  description = "The AWS account ID associated with the current caller identity"
+  description = "Deprecated. Use `aws_caller_identity.account_id`. The AWS account ID associated with the current caller identity"
   sensitive   = true
   value       = data.aws_caller_identity.this.account_id
 }
```
mjburling, line 29: Deprecating `account_id` to match how the platform module surfaces this. It might be interesting to start using the opentofu-specific `deprecation` fields once we're confident that all terraform is executed under opentofu....

PR #323, opened 2025-10-02
[PLT-1371] added workflows to dependabot

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1371

## 🛠 Changes

Add coverage for workflows in the .cdap dependabot configuration, including terraform.

## ℹ️ Context

Changes are for extended scan coverage by dependabot.  

## 🧪 Validation

see checks

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,5 +17,5 @@ updates:
   - package-ecosystem: terraform # Works for both Terraform and OpenTofu
     directory: /
     schedule:
-      interval: "daily"
+      interval: daily
     open-pull-requests-limit: 0


.github/dependabot.yml
```diff
@@ -1,15 +1,21 @@
 version: 2
 updates:
-  - package-ecosystem: github-actions
-    directory: /actions/aws-params-env-action/
+  - package-ecosystem: "github-actions"
```
gsf, line 3: String values don't need to be quoted in yaml

gsf (approved):

PR #392, opened 2026-02-12
WIP: Add SNS topic to Platform Module

mjburling:
## 🎫 Ticket


terraform/modules/platform/main.tf
```diff
@@ -157,3 +157,8 @@ data "aws_kms_alias" "secondary" {
 
   name = "alias/${local.app}-${local.parent_env}"
 }
+
+
+data "aws_sns_topic" "cloudwatch_alarms" {
+  name = "${local.app}-${local.parent_env}-cloudwatch-alarms"
+}
```
mianava, line 164: I don't think these are configured by default by CDAP at the moment so this could present issues with running the platform module and seeing no return. We could add the sns topic creation alongside ```module "sns_to_slack_queue" {``` and remove the establishment via the APIs, replacing with a platform module call

PR #437, opened 2026-04-09
PLT-1625 Service Connect 

mianava:
## 🎫 Ticket

## 🛠 Changes

This PR improves security posture and adds ECS Service Connect TLS support to the ECS service module. It also moves task role management and security group management into the module to reduce boilerplate in caller stacks.

- ssm:GetParameters now scoped to specific parameter ARNs; removed wildcard
- Task role moved into module; KMS always included and scoped to your vpc encryption (platform module derived) kms key;
- Cloudwatch log groups managed by Tofu with configurable retention (log_retention_days, default: 180) and KMS encryption
- Security group: Module creates and manages the task security group; callers add ingress/egress rules via task_security_group_id output
- Test stack : added services/tftesting/ecs-stack with two-service mTLS that can be validated with exec calls to the service-a, sets up a legacy ecs service to ensure backwards compatibility and non-enablement of service connect
- CI Workflow: adds github workflow to apply validate connection between and destroy a set of ecs services 

What is required change for the subscriber repo to upgrade to this version? 

<summary> 1. Import Your CloudWatch Log Group </summary>
<details>
The module now manages the CloudWatch log group. Before applying, import the existing one:
```
tofu import module.<your_module_name>.aws_cloudwatch_log_group.app \
  /aws/ecs/fargate/<app>-<env>/<service-name>
  ```
  
  On first apply, you'll see a non-destructive diff adding retention and KMS encryption. This is expected.
  </details>
  
<summary>  2. Remove Your External Task Role </summary>

<details>
The module now creates and manages the task role. Remove aws_iam_role.task, aws_iam_role_policy.task, and data.aws_iam_policy_document.task from your caller stack.

If your service needs app-specific permissions (S3, RDS, etc.), pass them as managed policy ARNs:

```
module "my_service" {
  additional_task_role_policies = {
    s3_read = aws_iam_policy.s3_read.arn
  }
}
```

Import the existing task role if the name matches the module's convention (<app>-<env>-<service>-task):

```
tofu import module.<your_module_name>.aws_iam_role.task <app>-<env>-<service>-task
```

</details>

<summary> 3. Remove Your External Security Group (If Applicable) </summary> 
<details> 
The module now creates and manages the ECS task security group. If you were previously creating a security group and passing it via security_groups, you can remove it.

The module creates the security group automatically and exposes its ID via the task_security_group_id output. Use this output to add ingress and egress rules in your caller stack:
```
# Allow inbound from ALB
resource "aws_vpc_security_group_ingress_rule" "alb_to_service" {
  security_group_id            = module.my_service.task_security_group_id
  from_port                    = 80
  to_port                      = 80
  ip_protocol                  = "tcp"
  referenced_security_group_id = aws_security_group.alb.id
  description                  = "Allow inbound HTTP from ALB"
}
```
import: 
```
tofu import module.<your_module_name>.aws_security_group.task <sg-id>
```
</details> 

What *can* you do with this version update? 
<summary> enable ecs service connect for mTLS</summary>
<details> 

1.    Create a Cloud Map namespace with a CA-permitted domain suffix (.cmscloud.local, .cms.local, etc.)
    Pass the namespace object to the module:
```
module "my_service" {
  enable_ecs_service_connect = true
  service_connect_namespace  = aws_service_discovery_http_namespace.this
}
```
The service module validates the namespace domain at plan time and will error with a clear message if the suffix isn't permitted by the Private CA. Default to cmscloud.local. 

2. Update to add a security group ingress rule on port 80 (service connect default) to the service that will be accessed via service connect (eg. ingress into worker from api) 

3. Update to add a security group egress rule on port 80 (service connect default) to the service that will be accessing via service connect (eg. egress from api into worker) 

4. Add relevant variables for sc_port and name, allowing you to map whatever the port your service is running on in your container to a friendly name. Your service can then call "worker" from "api" and doesn't require the port information or a DNS entry.

</details>

<summary> No changes are required for </summary>

1. Callers using the legacy load_balancers interface: dynamic "load_balancer" block now uses a single conditional block driven by local.enable_alb_integration rather than iterating over an external var.load_balancers list, var.load_balancers remains supported during transition 

2. service_name_override is retained; local.service_name resolves to var.service_name_override ?? var.platform.service, and local.service_name_full is used for all resource naming

3. container_secrets format, even though permissions are refined

## ℹ️ Context

These changes are made to support mTLS 

## 🧪 Validation

A new integration test stack (temporarily) lives at terraform/services/tftesting/ecs-stack/. It provisions two ECS services (tftesting-a and tftesting-b) in the same Service Connect namespace and validates that:
- Both services reach a healthy steady state
- service-a can successfully call service-b via Service Connect mTLS (curl http://tftesting-b:80/ returns HTTP 200)
- Envoy TLS handshake counters confirm mTLS is active

### CODE REVIEW

**Review by juliareynolds-nava — DISMISSED** [review]

File: terraform/modules/service/iam.tf [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+      "ecr:GetAuthorizationToken",
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:BatchGetImage",
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+      "ssm:GetParameters"
+    ]
+    resources = ["*"]
```
mianava: Refine this resources block for which ssm parameters are accessible to this function

jasonvinson (approved):

juliareynolds-nava (approved):

jscott-nava (approved):
LGTM! I'll have to make a few minor updates in order to be able to move the DPC services over onto this (i.e. output updates, etc.), but those would be more easily handled by a subsequent PR.

PR #438, opened 2026-04-09
PLT-1625 : ALB Module

mianava:
## 🎫 Ticket

## 🛠 Changes

Added a new standalone modules/alb Terraform module that provisions an ALB with a TLS-terminated HTTPS listener. This module is designed to be consumed by the ecs-service module (via alb_listener_arn) and the upcoming ecs-platform bundle.

Exposes a shared HTTPS listener with a 404 fixed-response default action, so multiple ECS services can each attach their own listener rules without owning the common ALB. 

Enables by default an optional HTTP to HTTPS redirect. 80 to 443.

## ℹ️ Context

This work is part of the broader effort to migrate services to use TLS everywhere and ECS Service Connect for mTLS, teams need a standardized way to provision an ALB with a valid certificate attached. 

Previously, ALB configuration was managed ad-hoc per team, leading to inconsistent policies.


mianava:
Closing in favor of another branch where this was merged in tandem with more tests.

PR #446, opened 2026-04-17
PLT-1652 : Datadog Dashboard Module

jasonvinson:
## 🎫 Ticket
https://jira.cms.gov/browse/PLT-1652

## 🛠 Changes
Added new terraform/modules/datadog_dashboard module with overview metrics for various services: Lambda, Aurora, SNS, S3, ECS, ELB. Teams can opt-out of widgets that are not relevant to their app.

## ℹ️ Context
This work is part of our migration from New Relic to Datadog. The goal is to provide a baseline dashboard that each team can quickly spin up while also enabling them to easily extend with custom widgets.

## 🧪 Validation
Validated by creating services/tftesting/datadog_dashboard and verifying the dashboard is setup correctly in the Datadog GUI after local tofu apply.


juliareynolds-nava (dismissed):

terraform/modules/datadog_dashboard/main.tf
```diff
@@ -0,0 +1,269 @@
...
+        }
+      }
+    }
+  }
+  dynamic "widget" {
+    # Renders the block exactly once if true, zero times if false
+    for_each = var.enable_default_widgets.elb ? [1] : []
+    content {
+      group_definition {
+        title       = "ELB"
```
mianava: Nit: This is logical because AWS is "ELB" for all the types, for consistency with how our services are, we might want to call this ALB.

terraform/modules/datadog_dashboard/main.tf
```diff
@@ -0,0 +1,269 @@
...
+        title = "Aurora Estimated Shared Memory (Bytes)"
+        request {
+          q = "avg:aws.rds.aurora_estimated_shared_memory_bytes{application:${var.app}, $env} by {environment}"
+        }
+      }
+    }
+
+
+  }
+}
```
mianava: Nit: Update PR with a line for all files. (Noting CDAP can add a pre-commit line for end of file in another PR.)

terraform/modules/datadog_dashboard/README.md
```diff
@@ -0,0 +1,35 @@
+<!-- BEGIN_TF_DOCS -->
```
mianava, line 6: Usability: It may be obvious, but a description here could help team members understand when they'd use this resource. In the status quo, the default setup should get them coverage for a common DASG application. We could also indicate that this module can be extended with an alternative default if folks are instantiating a dashboard that may be common but different than this in the future.

terraform/services/tftesting/datadog_dashboard/versions.tf
```diff
@@ -0,0 +1,15 @@
...
+  required_providers {
+    datadog = {
+      source  = "DataDog/datadog"
+      version = "~> 3.0"
+    }
+  }
+
+}
+
+provider "datadog" {
```
mianava, line 13: We can merge this in and update this when the application key path is confirmed in the API key and application key stuff. I *think* these can come from ssm parameters, so you'd, ideally, have a data lookup for the ssm parameter value and that will require the AWS provider, as well. Though I think we can keep that *out of* the modules/datadog_dashboard and keep it in the services/tftesting/datadog_dashboard where providers is defined.

terraform/services/tftesting/datadog_dashboard/versions.tf
```diff
@@ -0,0 +1,15 @@
...
+  api_url = "https://app.ddog-gov.com"
+}
```
mianava: Nit: line end formatting

mianava (changes requested):
Looks good! Just a few very small items. 

Let's step through the dashboard together to ensure redundancy of knowledge and get it merged in!

mianava (approved):
Approving though we can also wait to get the provider keys merged in in the other PR.

juliareynolds-nava (approved):

PR #452, opened 2026-04-22
PLT-1663 Allow ephemeral environment usage of firewall.

mianava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1663

## 🛠 Changes

Introduces the use of waf in ephemeral prod environment for ab2d
## ℹ️ Context

ab2d has issued an ephemeral environment in the prod vpc and that needs to have protections like production

## 🧪 Validation
This module has been validated in existing instantiations and will now be able to be instantiated by downstream repositories.


terraform/modules/firewall/variables.tf
```diff
@@ -8,11 +8,11 @@ variable "app" {
 }
 
 variable "env" {
-  description = "The application environment (dev, test, sbx, sandbox, prod)"
+  description = "The application environment (dev, test, sbx, sandbox, prod, ephemeral-prod)"
   type        = string
   validation {
-    condition     = contains(["dev", "test", "sbx", "sandbox", "prod"], var.env)
-    error_message = "Valid value for env is dev, test, sbx, sandbox, or prod."
+    condition     = contains(["dev", "test", "sbx", "sandbox", "prod", "ephemeral-prod"], var.env)
```
mjburling, line 14: This may be over-fit. Unless new restrictions have come up, valid ephemera are a little more forgiving. The loose definition has the name abiding by the terraform/tofu workspace naming conventions (lower case letters, numbers, and hyphens) **and** it must have a match for the regex `(dev|test|sbx|prod)$`.

For BFD, it's convention to use a number that signifies a Jira issue that the ephemeral environment supports as the prefix, e.g. the environment `9999-prod` would be a prod-seeded environment based off of BFD-9999. We do this to leave a something of a breadcrumb... it helps us identify who is responsible for the environment, if it's still in-use, etc.

I guess `ephemeral-prod` would be less of an ephemeral environment, and... just another established environment that's only used from time-to-time? ¯\\\_(ツ)_/¯

jasonvinson (approved):

juliareynolds-nava (approved):

PR #455, opened 2026-04-24
PLT-1662 Run log retention workflow with codebuild default python 

mianava:
## 🎫 Ticket

PLT-1662

## 🛠 Changes

Leverages the codebuild runner's default python version to run the cloudwatch retention script. 
Allows the log retention script to be run from a branch against any chosen environment. This enables testing and validation. 

## ℹ️ Context

These changes were made to ensure operability of the cloudwatch retention script via github workflows which was not yet operable. 

## 🧪 Validation
This was validated by running until the validation checks. Before merging, we will run the apply on production cloudwatch log groups with approval of the team overseeing compliance.

### DIFF
--- a/.github/workflows/set-log-retention.yml
+++ b/.github/workflows/set-log-retention.yml
@@ -106,7 +106,36 @@ jobs:
           echo "**Download the \`retention-plan\` artifact from the bottom of this page before approving.**" >> "$GITHUB_STEP_SUMMARY"
           
 
-  # ── JOB 2: APPROVE (manual gate) ───────────────────────────────────────────
+  # ── JOB 2 & 3: NOTIFY & APPROVE (manual gate) ───────────────────────────────────────────
+  notify:
+    name: "Pending Approval – Review Plan Details Here"
+    needs: plan
+    if: needs.plan.outputs.has_changes == 'true'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Post plan summary for reviewer
+        run: |
+          echo "## ⏳ Retention Plan Awaiting Approval"            >> "$GITHUB_STEP_SUMMARY"
+          echo ""                                                   >> "$GITHUB_STEP_SUMMARY"
+          echo "| Key | Value |"                                   >> "$GITHUB_STEP_SUMMARY"
+          echo "| --- | ----- |"                                   >> "$GITHUB_STEP_SUMMARY"
+          echo "| Plan file    | \`${{ needs.plan.outputs.plan_file }}\` |"       >> "$GITHUB_STEP_SUMMARY"
+          echo "| Target env   | \`${{ needs.plan.outputs.target_env }}\` |"      >> "$GITHUB_STEP_SUMMARY"
+          echo "| Generated at | \`${{ needs.plan.outputs.generated_at }}\` |"   >> "$GITHUB_STEP_SUMMARY"
+          echo "| Commands     | **${{ needs.plan.outputs.command_count }}** |"  >> "$GITHUB_STEP_SUMMARY"
+          echo ""                                                   >> "$GITHUB_STEP_SUMMARY"
+          echo "---"                                               >> "$GITHUB_STEP_SUMMARY"
+          echo "**Scroll and download the \`retention-plan\` artifact before approving.**" >> "$GITHUB_STEP_SUMMARY"
+          echo ""                                                   >> "$GITHUB_STEP_SUMMARY"
+          echo " Once reviewed, click **Review deployments** to approve or reject." >> "$GITHUB_STEP_SUMMARY"
+
+          # Also print to console log so it's visible in raw step output
+          echo "Plan file:    ${{ needs.plan.outputs.plan_file }}"
+          echo "Target env:   ${{ needs.plan.outputs.target_env }}"
+          echo "Generated at: ${{ needs.plan.outputs.generated_at }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/set-log-retention.yml
```diff
@@ -11,7 +10,7 @@ on:
         required: false
         default: "180"
       target_env:
-        description: "Target environment string (e.g. prod, test, dev, sandbox)"
+        description: "Target environment string (e.g. ephemeral-prod, happy-test, nice-sandbox, dev)"
         required: true
         default: "prod"
```
jadudm: Is it good for scripts to default to "prod?" Is this only ever run in prod? If not, should it be defaulted to a value that does not work, or is a lower/safe environment, in case the flow misbehaves?

E.g. default to "TARGET_ENV_NOT_SET" as opposed to "prod"?

.github/workflows/set-log-retention.yml
```diff
@@ -21,56 +20,108 @@ permissions:
...
   plan:
-    name: "Plan – Generate Retention Plan"
-    runs-on: codebuild-cdap-${{ github.ref_name == 'main' && 'prod' || 'non-prod' }}-${{ github.run_id }}-${{ github.run_attempt }}
+    name: "Plan - Generate Retention Plan"
+    runs-on: >-
+      codebuild-cdap-${{
+        (github.event_name == 'schedule'
+          || contains(github.event.inputs.target_env, 'prod')
+          || contains(github.event.inputs.target_env, 'sandbox'))
+        && 'prod'
```
jadudm, line 35: What is the order of operations of 

A && B || C

in the above expression? Does it guarantee that A is chosen first, and if that is not true, we necessarily get B? Are we guaranteed left-to-right evaluation on boolean expressions in this environment? (I really don't know.)

Parentheses might provide guarantees, if it matters?

That is, should this be

(A && B) || C 

or 

A && (B || C)

?

.github/workflows/set-log-retention.yml
```diff
@@ -21,56 +20,108 @@ permissions:
...
+      target_env:    ${{ steps.meta.outputs.target_env }}
+      generated_at:  ${{ steps.meta.outputs.generated_at }}
+      runner_env:    ${{ steps.resolve.outputs.runner_env }}
 
     steps:
+      - name: Resolve environment tier
+        id: resolve
+        run: |
+          if [ "${{ github.event_name }}" == "schedule" ]; then
+            echo "runner_env=prod" >> "$GITHUB_OUTPUT"
```
jadudm, line 53: Should we explicitly exit 0 here? Or, is it assumed that the only error pathway is where we exit 1, and everything else is an acceptable outcome?

.github/workflows/set-log-retention.yml
```diff
@@ -21,56 +20,108 @@ permissions:
...
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v4.3.0
 
-      - name: Set up Python ${{ env.PYTHON_VERSION }}
-        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v5.2.0
-        with:
-          python-version: ${{ env.PYTHON_VERSION }}
+      - name: Check available Python
+        working-directory: ${{ github.workspace }}
+        run: |
+          python3 --version
```
jadudm: Is this just to output a value to the environment? This step is essentially a no-op, because it does not assert anything. Is it just so we see the path and version in the logs?

.github/workflows/set-log-retention.yml
```diff
@@ -21,56 +20,108 @@ permissions:
...
 
       - name: Collect metadata
         id: meta
+        working-directory: ${{ github.workspace }}/scripts/set_log_retention
         run: |
-          PLAN_FILE=$(ls retention-plan-*.json | head -n 1)
+          PLAN_FILE=$(ls retention-plan-*.json 2>/dev/null | head -n 1)
+
+          if [ -z "$PLAN_FILE" ]; then
+            echo "No retention plan file found in $(pwd)"
```
jadudm: Consistency question: should `pwd` be `PWD`?

.github/workflows/set-log-retention.yml
```diff
@@ -85,82 +136,113 @@ jobs:
...
-  # ── JOB 3: APPLY ───────────────────────────────────────────────────────────
+  # -- JOB 4: APPLY -----------------------------------------------------------
   apply:
-    name: "Apply – Execute Retention Policy Updates"
+    name: "Apply - Execute Retention Policy Updates"
     needs: [plan, approve]
     if: needs.plan.outputs.has_changes == 'true'
-    runs-on: codebuild-cdap-${{ github.ref_name == 'main' && 'prod' || 'non-prod' }}-${{ github.run_id }}-${{ github.run_attempt }}
+    runs-on: >-
+      codebuild-cdap-${{
```
jadudm: Same boolean question as above. Probably occurs in multiple places.

.github/workflows/set-log-retention.yml
mianava: I suppose it would serve for debugging, though at this point it does seem irrelevant. I had originally neglected that Python was on the codebuild runners by default, and was doing an install. I switched to check the version and in some ways this remains from that process. Removing.

jasonvinson (approved):

jadudm (approved):
I had some questions that were resolved via live conversation. None of them were showstopper questions to begin with. Approving in that those questions were resolved/answered.

PR #456, opened 2026-04-25
PLT-1666 : Update lambda function module to support deployment and testing via Tofu

mianava:
## 🎫 Ticket

PLT-1666

### Alarm-to-slack function changes: 
- updates tags in alignment with standards module 
- builds function from source directory 
- turns on testing 
 
 
###  Lambda function changes: 
**The function module no longer includes SQS or broad SSM permissions in its default policy. Callers must now pass these in explicitly via:**

    ssm_parameter_paths — for SSM parameter access
    function_role_inline_policies — for SQS, S3, or other runtime permissions

- Logging: IAM restricted to cloudwatch log group for function, which is now managed to support common retention requirements: 
** Callers must now import existing cloudwatch log groups for management by the function module ** 

###  Lambda function testing: 
- creates a tftesting/function path that can be leveraged for local testing of Tofu changes to the function modules 
- creates a git workflow to evaluate the successful creation and destruction of a lambda function in the cdap test environment upon PR to the function module, allowing validation of CICD flows through Tofu for functions

Security group modifications were not made as part of these changes. 

## ℹ️ Context

These changes were motivated by updating the alarm-to-slack lambda predictably from source code in the CDAP repository. Additional enhancements were made to ensure IAM is refined for this function, which uplifted needs to refine the IAM broadly, enhance testablity, and ensure delivery pipeline predictability. 
 

## 🧪 Validation
After initial review, these changes will be deployed to cdap-test and a test alarm will be sent via SNS topic for evaluation.

<summary> bcda-test </summary>
<details>
Changes will be applied to bcda-test as in the git workflow plan output. 
1. A test alarm will be sent via AB2D's sns topic
2. A test alarm will be sent via DPC's sns topic

</details>

<summary> cdap-test </summary>
<details>

Changes will be applied to cdap-test as in the git workflow plan output.
1. A test alarm will be sent via DPC's sns topic. 

</details>


terraform/modules/function/iam.tf
```diff
@@ -0,0 +1,156 @@
...
+  }
+}
+
+resource "aws_iam_role" "function" {
+  name = "${var.name}-function"
+  path = "/delegatedadmin/developer/"
+
+  permissions_boundary = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:policy/cms-cloud-admin/developer-boundary-policy"
+
+  assume_role_policy = data.aws_iam_policy_document.function_assume_role.json
```
mianava, line 122: I need to dig into if this is still relevant in our permissions setup.

terraform/modules/function/main.tf
```diff
@@ -1,187 +1,48 @@
...
-    bcda = [
-      "repo:CMSgov/bcda-app:*",
-    ]
-    cdap = [
-      "repo:CMSgov/cdap:*",
-    ]
-    dpc = [
-      "repo:CMSgov/dpc-app:*",
-    ]
-  }
```
mianava, line 14: These are moved into a variable in the case that repositories are managing deployment. See variable `github_actions_repos`

terraform/modules/function/main.tf
```diff
@@ -1,187 +1,48 @@
...
 
-  name   = each.key
-  role   = aws_iam_role.function.id
-  policy = each.value
-}
-
-data "aws_ssm_parameter" "prod_account_id" {
-  count = var.env == "test" ? 1 : 0
-  name  = "/prod/account-id"
-}
```
mianava, line 142: `cdap-mgmt` is being deprecated in favor of a cdap- vpc managing lambdas in a dedicated fashion within accounts: cross-account access is no longer relevant.

terraform/modules/function/variables.tf
```diff
@@ -60,44 +62,93 @@ variable "memory_size" {
...
+  default     = {}
+}
+
+# ── Advanced / Migration strategies ─────────────────────────────────────────────────
+
+variable "extra_kms_key_arns" {
   description = "Optional list of additional KMS key ARNs the Lambda can use"
+  type        = list(string)
+  default     = []
 }
```
mianava, line 171: Is this used anywhere? Do we have a use case for this?

terraform/modules/function/main.tf
```diff
@@ -273,3 +131,11 @@ resource "aws_lambda_permission" "cloudwatch_events" {
   principal     = "events.amazonaws.com"
   source_arn    = aws_cloudwatch_event_rule.this[0].arn
 }
+
+# Manage cloudwatch log group to ensure compliant
```
juliareynolds-nava, line 118: This is going to be problematic for consumer teams.  Every time after the first time that we are iteratively testing a new lambda deployment in a lower environment we will have to manually delete the log group and if it has skip destroy it will be un-deletable.  Suggest you remove the cloudwatch log group and let teams handle it, especially since if it's not defined lambda will create its own log group.  Your plan to script cloudwatch retention settings in prod is a better idea.

mianava: Ooh good catch. I updated this so that it's only retained when the env string includes `prod`. The retention_days setting, though, I left the default to 180 days which may be excessive for non-prod environments.

terraform/services/alarm-to-slack/main.tf
```diff
@@ -2,22 +2,46 @@ locals {
...
+  }
+
+  # Point to the local source directory — module handles zip + upload
+  source_dir = "${path.module}/lambda_src"
+
+  # Optionally exclude tests and cache
+  source_dir_excludes = [
+    "__pycache__",
+    "test_*.py",
+    "*.pyc",
```
lukey-luke, line 42: 👍

terraform/modules/function/iam.tf
```diff
@@ -0,0 +1,156 @@
...
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "Service"
+      identifiers = ["lambda.amazonaws.com"]
+    }
+  }
+
+  # Allow access from GitHub-hosted runners via OIDC for integration tests
+  dynamic "statement" {
```
lukey-luke, line 31: Are these changes covered in tests somewhere in the cdap repo? Any additional workflow changes required?

I reckon this might be a larger effort and should be moved to a separate PR, but please let me know where I can review

terraform/modules/function/variables.tf
```diff
@@ -60,44 +62,125 @@ variable "memory_size" {
...
   description = "Cron or rate expression for a scheduled function"
   type        = string
   default     = ""
 }
 
-variable "extra_kms_key_arns" {
+variable "log_retention_days" {
+  description = "Number of days to retain Lambda function logs in CloudWatch. If null, no retention policy is set and retention is managed externally (e.g., via cdap/scripts/set_log_retention/)."
+  type        = number
+  default     = 180
```
lukey-luke, line 136: If we already have a mechanism to manage logs and there's a program-wide standard for retention period, it might be easier to make this optional/default to null.

Then default log retention period for CW logs only lives in one place. This way, if default needs to change from 180 days to 90 days for example, there are less moving pieces.

lukey-luke:
I verified the webhook_urls already exist for DPC on parameter store ✅ 
File structure for `alarm-to-slack/` makes sense and follow existing conventions ✅ 
Python code is up-to-date w/ arm64 updates ✅ 
Active thread on CW logs 🔁

Changes in function module for CI/CD enablement are not clear to me ❌ 
- Consider moving to a separate ticket or PR 
- Please include additional links or info on how to make use of `github_actions_repos`

terraform/modules/function/iam.tf
mianava: I updated this PR with a test function that will also come up and down on apply.

juliareynolds-nava (dismissed):

terraform/modules/function/variables.tf
mianava: To support standard encryption and retention, setting this via terraform will be the most straightforward way to ensure no log groups fall through the cracks. CMS is setting the default at 30 days, and if we run a check every week to bump cloudwatch log groups to 180 days, we risk missing log groups that are provisioned for a day or two and updated and not updating their retention. 

We can, however, use a standard cloudwatch_log_group module and call that from other modules. This can be updated in a future PR for all the modules.

jasonvinson (dismissed):

juliareynolds-nava (dismissed):

.github/workflows/alarm-to-slack-checks.yml
mianava, line 1: This is currently failing and the view is not visible in sonarqube. Removing until https://jiraent.cms.gov/plugins/servlet/desk/portal/22/CLDSPT-107117 is resolved.

terraform/modules/function/iam.tf
mianava: Included in a ticket https://jira.cms.gov/browse/PLT-1667

juliareynolds-nava (approved):

jasonvinson (approved):

PR #457, opened 2026-04-29
PLT-1651 : Add DataDog agent sidecar to ECS service module

jasonvinson:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1651

## 🛠 Changes

Updated ECS Service module main.tf to include a Datadog agent container that is launched alongside the main application container. Logging is disabled as it's not approved by CMS.

## ℹ️ Context

New Relic is being deprecated. CMS has provided DataDog as its replacement.

## 🧪 Validation

Validated by launching a test cluster and verifying the agent sidecar runs alongside the main app container. Data reported by agent is visible in DataDog GUI.


juliareynolds-nava (approved):

terraform/modules/service/main.tf
```diff
@@ -1,6 +1,40 @@
...
+        awslogs-region        = var.platform.primary_region.name
+        awslogs-stream-prefix = "${var.platform.app}-${var.platform.env}"
+      }
+    }
+    healthCheck = var.health_check
+  }
+
+  datadog_container = {
+    name      = "datadog-agent"
+    image     = "public.ecr.aws/datadog/agent:latest"
```
mianava: Can this version be pinned?

terraform/modules/service/main.tf
```diff
@@ -1,6 +1,40 @@
...
+    essential = false # Do not impact task health if this container fails
+    environment = [
+      { name = "ECS_FARGATE", value = "true" },
+      { name = "DD_SITE", value = "ddog-gov.com" },
+      { name = "DD_APM_ENABLED", value = "true" },
+      { name = "DD_LOGS_ENABLED", value = "false" },  # DD logging is currently not approved
+      { name = "DD_ECS_TASK_COLLECTION_ENABLED", value = "true" }
+    ]
+    secrets = var.container_secrets
```
mianava: I don't think we want all secrets available to the Datadog agent container if they can be split. This should have limited access to just the API key needed. I think the whole policy can live within this module, also.

mianava (changes requested):

terraform/modules/service/main.tf
jasonvinson: Changed to secrets = [{ name = "DD_API_KEY", valueFrom = "/${var.platform.app}/${var.platform.env}/datadog/agents/api_key" }]

terraform/modules/service/main.tf
jasonvinson: Done

juliareynolds-nava (approved):

mianava (approved):

PR #458, opened 2026-04-30
PLT-1652 : Pull Datadog credentials from SSM

jasonvinson:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1652

## 🛠 Changes

- Adopt Standards module to pull Datadog credentials from SSM in the /tftesting/datadog_dashboard/ test module.
- Renamed versions.tf -> tofu.tf
 
## ℹ️ Context

Previously, our API creds were not available in SSM yet, so we were using local creds to test.

## 🧪 Validation

Performed tofu apply and verified that the dashboard is created successfully.


juliareynolds-nava (dismissed):

terraform/modules/datadog_dashboard/tofu.tf
```diff
@@ -0,0 +1,12 @@
+terraform {
+  required_providers {
+    datadog = {
+      source  = "DataDog/datadog"
+      version = "~>4.4"
+    }
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~>5"
```
mianava: Let's bump this to 6 as much as possible

terraform/services/tftesting/datadog_dashboard/main.tf
```diff
@@ -1,31 +1,42 @@
 locals {
-  app_name = "ab2d"
+  cdap_env = contains(["sandbox", "prod"], var.env) ? "prod" : "test"
```
mianava: I think you can grab this from account_env_suffix in standards module - if not, it's modeled in platform and that might be one way to get it predictably

terraform/services/tftesting/datadog_dashboard/tofu.tf
```diff
@@ -0,0 +1,37 @@
+terraform {
+  required_providers {
+    datadog = {
+      source  = "DataDog/datadog"
+      version = "~>4.4"
+    }
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~>5"
```
mianava: Bump to 6

terraform/services/tftesting/datadog_dashboard/variables.tf
```diff
@@ -0,0 +1,17 @@
+variable "app" {
+  description = "The application name (ab2d, bcda, cdap dpc)"
```
mianava: I think for this one we are going to want to allow bbapi, too. Might want to update throughout

terraform/services/tftesting/datadog_dashboard/variables.tf
```diff
@@ -0,0 +1,17 @@
...
+  type        = string
+  validation {
+    condition     = contains(["ab2d", "bcda", "cdap", "dpc"], var.app)
+    error_message = "Valid value for app is ab2d, bcda, cdap or dpc."
+  }
+}
+
+variable "env" {
```
mianava: I think for this one all of these won't be allowed. In theory, they're all "datadog prod" though I do appreciate that the dashboard for "tftesting" would be handled in "test".

mianava (changes requested):

terraform/services/tftesting/datadog_dashboard/main.tf
jasonvinson: Removed as it's unnecessary for this test app.

mianava (approved):

juliareynolds-nava (approved):

PR #461, opened 2026-05-05
BCDA-9967: add model short name

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9967

## 🛠 Changes

Adding `model_name` to the production and sandbox jobs view

## ℹ️ Context

We want to have the Alternative Payment Model included in job request data so that it can be easily aggregated, without having to derive the model from each entity's CMS ID

## 🧪 Validation

Executed queries against prod/sandbox locally


carlpartridge (approved):

laurenkrugen-navapbc (approved):

jscott-nava (approved):
Looks good.

PR #477, opened 2026-05-20
PLT-1672 : Datadog Lambda Instrumentation

jasonvinson:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1672

## 🛠 Changes

Adds a Datadog lambda layer to to modules/function module to allow reporting of enhanced metrics and APM.

## ℹ️ Context

NewRelic is being sunset with a deadline of June 30, 2026. To support teams tracking infrastructure metrics/APM, we are switching to Datadog which has been provided by CMS.

## 🧪 Validation

Validated by deploying function, verifying DD layer is present, and verifying enhanced metrics and traces are displayed for the lambda in Datadog console.

### DIFF
--- a/terraform/modules/function/main.tf
+++ b/terraform/modules/function/main.tf
@@ -58,7 +58,27 @@ module "subnets" {
   vpc_id = module.vpc.id
 }
 
-resource "aws_lambda_function" "this" {
+module "lambda-datadog" {
+  source  = "DataDog/lambda-datadog/aws"
+  version = "4.7.0"
+
+  environment_variables = merge(var.environment_variables, {
+    "DD_API_KEY_SSM_ARN"          : data.aws_ssm_parameter.dd_api_key.arn
+    "DD_ENV"                      : local.env
+    "DD_SERVICE"                  : local.app
+    "DD_SITE"                     : "ddog-gov.com"
+    "DD_VERSION"                  : var.source_code_version
+    "DD_SERVERLESS_LOGS_ENABLED"  : false
+    "DD_LAMBDA_HANDLER"           : var.handler
+  })
+
+  datadog_extension_layer_version = 96
+  datadog_python_layer_version = startswith(var.runtime, "python") ? var.dd_python_layer_version : null
+  datadog_node_layer_version   = startswith(var.runtime, "nodejs") ? var.dd_node_layer_version : null
+  datadog_java_layer_version   = startswith(var.runtime, "java") ? var.dd_java_layer_version : null
+  datadog_ruby_layer_version   = startswith(var.runtime, "ruby") ? var.dd_ruby_layer_version : null
+  datadog_dotnet_layer_version = startswith(var.runtime, "dotnet") ? var.dd_dotnet_layer_version : null
+
   description   = var.description
   function_name = local.full_name_string
   s3_key        = "function.zip"
@@ -76,18 +96,9 @@ resource "aws_lambda_function" "this" {
   layers        = var.layer_arns
   architectures = [var.architecture]
 
-  tracing_config {
-    mode = "Active"
-  }
-
-  vpc_config {
-    subnet_ids         = module.subnets.ids
-    security_group_ids = [aws_security_group.function.id]
-  }
-
-  environment {
-    variables = var.environment_variables
-  }
+  tracing_config_mode = "Active"
+  vpc_config_subnet_ids         = module.subnets.ids
+  vpc_config_security_group_ids = [aws_security_group.function.id]
 }
 
 resource "aws_cloudwatch_event_rule" "this" {
@@ -101,7 +112,7 @@ resource "aws_cloudwatch_event_rule" "this" {
 resource "aws_cloudwatch_event_target" "this" {
   count = var.schedule_expression != "" ? 1 : 0
 
-  arn  = aws_lambda_function.this.arn
+  arn  = module.lambda-datadog.arn
   rule = aws_cloudwatch_event_rule.this[0].name
 }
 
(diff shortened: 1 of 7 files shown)


juliareynolds-nava (dismissed):

mianava (changes requested):
Can we make this *optional*? Adding the lambda layer sets up APM for usage but I don't think is necessary for all lambdas. I *think* this could mean adding a flag like "enable_dd" and explaining the use-case, and pointing to additional work, and then putting count values where relevant.

terraform/modules/function/main.tf
```diff
@@ -58,7 +58,27 @@ module "subnets" {
...
+    "DD_API_KEY_SSM_ARN"          : data.aws_ssm_parameter.dd_api_key.arn
+    "DD_ENV"                      : local.env
+    "DD_SERVICE"                  : local.app
+    "DD_SITE"                     : "ddog-gov.com"
+    "DD_VERSION"                  : var.source_code_version
+    "DD_SERVERLESS_LOGS_ENABLED"  : false
+    "DD_LAMBDA_HANDLER"           : var.handler
+  })
+
+  datadog_extension_layer_version = 96
```
mianava: can we expose this as a variable so it's very obvious to us we have to increment the default?

jasonvinson:
> Can we make this _optional_? Adding the lambda layer sets up APM for usage but I don't think is necessary for all lambdas. I _think_ this could mean adding a flag like "enable_dd" and explaining the use-case, and pointing to additional work, and then putting count values where relevant.

I added a "dd_enabled" flag and updated the module to include both the datadog-wrapped version of the lambda and the original lambda itself. I used count logic to choose with resource to deploy based on the new flag.

I tested applying with both "dd_enabled" set to true and false in the tftesting function and verified the correct version of the lambda was deployed in both cases.

terraform/modules/function/main.tf
```diff
@@ -58,23 +84,62 @@ module "subnets" {
...
-  aws_s3_object.function_zip[0].version_id : var.source_code_version)
+module "lambda-datadog" {
+  count   = var.dd_enabled ? 1 : 0  # Only create if using Datadog Lambda module (var.dd_enabled = true)
+  source  = "DataDog/lambda-datadog/aws"
+  version = "4.7.0"
+
+  environment_variables = merge(var.environment_variables, {
+    "DD_API_KEY_SSM_ARN" : data.aws_ssm_parameter.dd_api_key.arn
+    "DD_ENV" : local.env
+    "DD_SERVICE" : local.app
```
mianava, line 95: I think this can map to function name or var.platform.service?

terraform/modules/function/main.tf
```diff
@@ -58,23 +84,62 @@ module "subnets" {
...
-  description   = var.description
-  function_name = local.full_name_string
-  s3_key        = "function.zip"
-  s3_bucket     = module.zip_bucket.id
-  # null = use latest S3 version; set = pin to a specific prior version
-  s3_object_version = var.rollback_version != null ? var.rollback_version : (var.source_dir != null ?
-  aws_s3_object.function_zip[0].version_id : var.source_code_version)
+module "lambda-datadog" {
+  count   = var.dd_enabled ? 1 : 0  # Only create if using Datadog Lambda module (var.dd_enabled = true)
+  source  = "DataDog/lambda-datadog/aws"
```
mianava, line 89: We aren't using external providers for Tofu -

mianava (changes requested):

juliareynolds-nava (approved):

terraform/modules/function/main.tf
```diff
@@ -4,6 +4,32 @@ locals {
   env             = var.platform.env
 
   full_name_string = "${local.app}-${local.env}-${var.name}"
+
+  # null = use latest S3 version; set = pin to a specific prior version
+  lambda_s3_object_version = var.rollback_version != null ? var.rollback_version : (var.source_dir != null ?
```
jadudm, line 9: Question: is it common practice to nest conditionals, or would it make sense to break this into two lines, with the `source_dir` conditional as its own value, and then use it in the `s3_object_version` conditional on the next line?

terraform/modules/function/main.tf
```diff
@@ -4,6 +4,32 @@ locals {
...
+    role              = aws_iam_role.function.arn
+    handler           = var.handler
+    runtime           = var.runtime
+    timeout           = var.timeout
+    memory_size       = var.memory_size
+    layers            = var.layer_arns
+    architectures     = [var.architecture]
+  }
+
+  lambda_function_arn              = var.dd_enabled ? module.lambda-datadog[0].arn : aws_lambda_function.this[0].arn
```
jadudm, line 28: A new-to-TF question:

Would this code be simplified if we did something like 

```
lam_fun = var.dd_enabled ? module.lambda-datadog[0] : aws_lambda_function.this[0]
```

and then use lam_fun in the rest of the expressions? Or, does that not work?

```
lambda_function_name = lam_fun.function_name
lambda_function_version = lam_fun.version
...
```

terraform/modules/function/main.tf
```diff
@@ -58,23 +84,62 @@ module "subnets" {
...
-  timeout       = var.timeout
-  memory_size   = var.memory_size
-  layers        = var.layer_arns
-  architectures = [var.architecture]
+  datadog_extension_layer_version = var.dd_extension_layer_version
+  datadog_python_layer_version    = startswith(var.runtime, "python") ? var.dd_python_layer_version : null
+  datadog_node_layer_version      = startswith(var.runtime, "nodejs") ? var.dd_node_layer_version : null
+  datadog_java_layer_version      = startswith(var.runtime, "java") ? var.dd_java_layer_version : null
+  datadog_ruby_layer_version      = startswith(var.runtime, "ruby") ? var.dd_ruby_layer_version : null
+  datadog_dotnet_layer_version    = startswith(var.runtime, "dotnet") ? var.dd_dotnet_layer_version : null
```
jadudm, line 107: Is there a `golang` version? Only asking because I see a list of languages/runtimes...

jasonvinson: golang requires "manual" installation of the specific lambda package: https://docs.datadoghq.com/serverless/aws_lambda/instrumentation/go/?tab=terraform

PR #482, opened 2026-05-27
PLT-1644 : Notifications to Slack and Victorops Webhooks 

mianava:
## 🎫 Ticket

PLT-1644 

## 🛠 Changes

- sets up notifications to write to slack based on var.app or not 
- sets up notification webhooks via datadog_webhooks 
- moves defaults.yml so it serves as an example not as a default exhaustive list that might cause conflict in the future 
- sets up custom monitors via YAML settings as an object vs as full tofu resources 

## 🧪 Validation

These were validated by deploying changes to the codebuild alerts that are custom alerts and verifying that the alert comes through to Slack for test env. The victorops integration was also set up and tested for alert throughput.


terraform/modules/datadog_monitors/config_example/example_defaults.yml
```diff
@@ -0,0 +1,187 @@
...
+# -----------------------------------------------------------------------------
+# Notifications
+#
+# The global platform channel is always included. Add your team's channels
+# in your environment config under notifications.channels.
+
+# -----------------------------------------------------------------------------
+notifications:
+  slack:
+    - "cdap-users"
```
mianava: Change this to the boolean

gfreeman-navapbc (dismissed):

jasonvinson (approved):

gfreeman-navapbc (approved):

PR #489, opened 2026-06-09
Switch from cmscloud.internal to internal.cms.gov. Establish hosted zone

jasonvinson:
Switch from cmscloud.internal to internal.cms.gov. Establish cdap hosted zone

## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1694

## 🛠 Changes

1. Update ACM Cert module to provision cert for internal.cms.gov instead of cmscloud.internal.
2. Update hosted zone service with internal.cms.gov also.

## ℹ️ Context

These changes will enable CDAP to host Datadog private locations to allow sythetic testing of API teams' private internal endpoints via Datadog Synthetics.

## 🧪 Validation

Changes will be validated after merge.

### DIFF
--- a/terraform/modules/acm_certificate/variables.tf
+++ b/terraform/modules/acm_certificate/variables.tf
@@ -19,9 +19,9 @@ variable "enable_internal_endpoint" {
   default     = false
   description = <<-EOT
     Issue a PCA-backed certificate for the VPC-internal endpoint.
-    Domain: <app>-<env>-<service>.internal.cms.gov
+    Domain: <service>.<env>.<app>.internal.cms.gov
     Use for Lambda/ECS-to-ECS calls that do not need Zscaler or public access.
-    Route 53 is NOT managed here — DNS for .internal.cms.gov is handled by CMS.
+    Route 53 is NOT managed here.
   EOT
 }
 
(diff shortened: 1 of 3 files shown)


terraform/modules/acm_certificate/variables.tf
```diff
@@ -12,16 +12,16 @@ variable "platform" {
...
-# Internal endpoint (VPC-only, cmscloud.internal)
+# Internal endpoint (VPC-only, internal.cms.gov)
 # -------------------------------------------------------
 variable "enable_internal_endpoint" {
   type        = bool
   default     = false
   description = <<-EOT
     Issue a PCA-backed certificate for the VPC-internal endpoint.
-    Domain: <app>-<env>-<service>.internal
+    Domain: <app>-<env>-<service>.internal.cms.gov
```
mianava: I think we might want this to be <service>.<env>.<app>.internal.cms.gov

Because the envs are in different accounts so the hosted zones will be, we'd end up registering each <env>.<app>... with CMS if we are requiring dns resolution.

terraform/modules/acm_certificate/variables.tf
```diff
@@ -12,16 +12,16 @@ variable "platform" {
...
     Use for Lambda/ECS-to-ECS calls that do not need Zscaler or public access.
-    Route 53 is NOT managed here — DNS for .internal is handled by CMS.
+    Route 53 is NOT managed here — DNS for .internal.cms.gov is handled by CMS.
```
mianava: Ah, this is distinct. We will be handling DNS for private endpoints, which internal.cms.gov is, though <app>.cms.gov will continue to be handled by CMS. Little confusing with the same domain ending.

mianava: Though Route53 is, indeed, not handled here this is just the certs at the moment.

mianava (dismissed):
Just requests to refine the comments.

terraform/modules/acm_certificate/variables.tf
jasonvinson: service.env.app.internal.cms.gov

mianava (approved):
LGTM to get moving on testing / evaluating this with the CDAP Datadog service that might use them.

juliareynolds-nava (approved):

PR #514, opened 2026-07-10
BB2-4951 Allow conditional creation of custom monitors

annamontare-nava:
## 🎫 Ticket

BB2-4951

## 🛠 Changes

Adds a create option to custom monitors.

By setting create to an expression, monitors can be, for example, only
created on certain environments.

## ℹ️ Context

Teams should probably use this sparingly, having monitors be the same across environments usually makes sense.


mianava:
Is it possible to favor this being set in config? default.yml if for all envs or in the <env>.yml for relevant environment? I suppose this is to make exclusion easier and therefore reduce code?

annamontare-nava:
> Is it possible to favor this being set in config? default.yml if for all envs or in the .yml for relevant environment? I suppose this is to make exclusion easier and therefore reduce code?

What are you envisioning would be in the config?

mianava (commented):
I'm still inclined to favor this being defined only where it's used, which I would assume would be only `prod` if it's disabled elsewhere. If it meaningfully reduces code volume to have this option, let's go for it, though I'm having a hard time understanding the long term use case at the moment.

mianava (dismissed):

mianava (approved):
Approving for sparing use, noting that only very specific monitors present an environment based opt-out option. Most monitors should exist in all environments.

PR #516, opened 2026-07-13
PLT-1694 : datadog private location

jasonvinson:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1694

## 🛠 Changes

- Add usage guide to README.md

## ℹ️ Context

Updating documentation.

## 🧪 Validation

N/A


terraform/services/520-datadog-private-location/README.md
```diff
@@ -1,7 +1,95 @@
...
+  ...
+  enable_datadog_synthetics_ingress = true
+}
+```
+
+**What this does under the hood:**
+- The service module reads the PL worker's security group ID from SSM at `/cdap/${env}/datadog/nonsensitive/private_location_task_security_group_id`
+- It creates an `aws_vpc_security_group_ingress_rule` on your task's SG that references the PL's SG, allowing all inbound traffic from the PL runner
+
+> **Prerequisite:** Your service must be in a VPC already listed in the PL's config file (terraform/services/520-datadog-private-location/config/). If your VPC is not listed, open a PR to add it.
```
mianava: What's also implicit here is, I believe, that the VPC is peered with CDAP's for private network traffic.

terraform/services/520-datadog-private-location/README.md
```diff
@@ -1,7 +1,95 @@
...
+  name    = "your-app-${var.env}-api-health"
+  type    = "api"
+  subtype = "http"
+  status  = "live"
+
+  request_definition {
+    method = "GET"
+    url    = "http://your-internal-alb-dns/health"
+  }
+
```
mianava: Can this example be moved to use a config/ example? This might be easiest if evaluated and tested with a reusable module, though a dedicated terraservice in cdap in the #600 level might make sense, too. The synthetic in cdap could ping an external public endpoint.

mianava (changes requested):

terraform/services/520-datadog-private-location/README.md
jasonvinson: Changed to use new synthetics module.

terraform/services/520-datadog-private-location/README.md
```diff
@@ -1,7 +1,58 @@
...
+---
+
+### Step 2 — Write the Synthetic Test in Terraform
+
+Use the `datadog_synthetics_test` resource. Look up the private location ID by name using `data.datadog_synthetics_locations`:
+
+See terraform/services/tftesting/datadog-synthetic for example.
+
+The display name prefix for each environment:
```
mianava, line 35: Is there a test example of this? Do we have a synthetic running through this private location that was set up with Datadog?

terraform/services/520-datadog-private-location/README.md
```diff
@@ -1,7 +1,58 @@
...
+| `test` | `cdap-non-prod` |
+| `prod` | `cdap-prod` |
+
+---
+
+### Step 3 — Wire it to a Monitor (Optional but Recommended)
+
+Once the synthetic test is created, attach a `datadog_monitor` of type `"synthetics alert"` referencing the test's ID to get alerting via the existing Datadog → Slack pipeline.
```
mianava, line 43: These should be set up as custom_monitors, preferably. Can that be modeled out possibly in cdap's monitors service? This could be with a simple dummy test

mianava (changes requested):

terraform/services/520-datadog-private-location/README.md
jasonvinson: terraform/services/tftesting/datadog-synthetic/main.tf shows resolving and running a synthetic through the PL. Can also see test results in the DD console: cdap-test-private-location-connectivity

mianava (approved):
LGTM, noting we may change the number of failures allowed to be more than 0 if we find this is noisy at all.

PR #517, opened 2026-07-13
PLT-1717 : Explicitly manage Log Group for ECS Container Insights

jasonvinson:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1717

## 🛠 Changes

terraform/modules/cluster/main.tf: added new log group ecs_container_insights_logs

## ℹ️ Context

We are required to retain logs for 180 days. Since Container Insights are enabled, if we don't manage this log group explicitly, it will be created automatically by AWS and we won't be able to set the retention period via Terraform.

## 🧪 Validation

Applied to test env and verified retention period is correctly set.

### DIFF
--- a/terraform/modules/cluster/main.tf
+++ b/terraform/modules/cluster/main.tf
@@ -1,10 +1,31 @@
+locals {
+  cluster_name = var.cluster_name_override != null ? var.cluster_name_override : "${var.platform.app}-${var.platform.env}-${var.platform.service}"
+}
+
+# CloudWatch Log Group for ECS Container Insights. If we don't manage this explicitly, it will be created automatically by AWS and we won't be able to manage the retention period via Terraform.
+resource "aws_cloudwatch_log_group" "ecs_container_insights_logs" {
+  name              = "/aws/ecs/containerinsights/${local.cluster_name}/performance"
+  retention_in_days = 180
+  kms_key_id        = var.platform.kms_alias_primary.target_key_arn
+
+  tags = {
+    Application = var.platform.app
+    Environment = var.platform.env
+    Service     = var.platform.service
+  }
+}
+
 resource "aws_ecs_cluster" "this" {
-  name = var.cluster_name_override != null ? var.cluster_name_override : "${var.platform.app}-${var.platform.env}-${var.platform.service}"
+  name = local.cluster_name
 
   setting {
     name  = "containerInsights"
     value = "enabled"
   }
+
+  depends_on = [
+    aws_cloudwatch_log_group.ecs_container_insights_logs
+  ]
 
   configuration {
     managed_storage_configuration {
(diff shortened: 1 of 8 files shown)


terraform/modules/cluster/main.tf
```diff
@@ -1,11 +1,32 @@
+locals {
+  cluster_name = var.cluster_name_override != null ? var.cluster_name_override : "${var.platform.app}-${var.platform.env}-${var.platform.service}"
+}
+
+# CloudWatch Log Group for ECS Container Insights. If we don't manage this explicitly, it will be created automatically by AWS and we won't be able to manage the retention period via Terraform.
+resource "aws_cloudwatch_log_group" "ecs_container_insights_logs" {
+  name              = "/aws/ecs/containerinsights/${local.cluster_name}/performance"
+  retention_in_days = 180
+  kms_key_id        = var.platform.kms_alias_primary.target_key_arn
```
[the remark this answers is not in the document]
jasonvinson: Fixed.

terraform/modules/cluster/main.tf
```diff
@@ -1,11 +1,32 @@
...
+
+  tags = {
+    Application = var.platform.app
+    Environment = var.platform.env
+    Service     = var.platform.service
+  }
+}
```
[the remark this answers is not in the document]
jasonvinson: Fixed.

terraform/modules/cluster/main.tf
```diff
@@ -1,11 +1,32 @@
...
+resource "aws_cloudwatch_log_group" "ecs_container_insights_logs" {
+  name              = "/aws/ecs/containerinsights/${local.cluster_name}/performance"
+  retention_in_days = var.log_retention_days
+  kms_key_id        = var.platform.kms_alias_primary.target_key_arn
+
+  tags = {
+    Application = var.platform.app
+    Environment = var.platform.env
+    Service     = var.platform.service
+  }
```
mianava: These are not needed as they should come from default tags.

terraform/modules/cluster/variables.tf
```diff
@@ -16,3 +16,18 @@ variable "cluster_name_override" {
...
+
+  validation {
+    condition = contains(
+      [0, 1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731,
+      1096, 1827, 2192, 2557, 2922, 3288, 3653],
+      var.log_retention_days
+    )
+    error_message = "log_retention_days must be a value supported by CloudWatch Logs (e.g. 30, 90, 180, 365, 731). See: https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutRetentionPolicy.html"
+  }
+}
```
mianava, line 24: It might be more efficient to establish a "cloudwatch_log_group" module and move this there, and log a ticket to migrate the other cloudwatch_log_group calls there, too. While that does create nesting of modules, it will meaningfully reduce the replication here and make it easier to set the log retention confidently.

mianava (changes requested):

terraform/modules/cluster/variables.tf
jasonvinson: Good idea! Done :)

mianava (approved):

PR #522, opened 2026-07-15
BB2-4951 add avg time per request graph to shared dashboard

annamontare-nava:
## 🎫 Ticket

BB2-4951

## 🛠 Changes

Add a breakdown of average time per request to the shared dashboard.

Might need to update this to reflect the work in #518. One question then is, does the graph still make sense if it is filtered by application rather than by service?

## ℹ️ Context

This graph is visible on the APM service page, and breaks down average
time spent per request by service / its dependencies.

This is one of the graphs the Blue Button team had in New Relic, so replicating it here in the dashboard.

Counter / question: it's already on the service page, so is having it on the dashboard helpful?

## 🧪 Validation

needs validation, can deploy to bb dashboard and see if it works there

### DIFF
--- a/terraform/modules/datadog_dashboard/main.tf
+++ b/terraform/modules/datadog_dashboard/main.tf
@@ -183,6 +183,38 @@ resource "datadog_dashboard" "application_metrics_dashboard" {
 
         widget {
           timeseries_definition {
+            title = "Avg Time per Request"
+            live_span = var.widget_live_spans.apm
+            request {
+              display_type = "area"
+              query {
+                metric_query {
+                  name = "query1"
+                  query = "sum:trace.${var.apm_primary_operation}.exec_time.by_service{service:${var.app}, $env} by {sublayer_service, sublayer_inferred}.rollup(sum).fill(zero)"
+                }
+              }
+              query {
+                metric_query {
+                  name = "query2"
+                  query = "sum:trace.${var.apm_primary_operation}.hits{service:${var.app}, $env}.rollup(sum).fill(zero).as_count()"
+                }
+              }
+              formula {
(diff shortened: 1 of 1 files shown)


terraform/modules/datadog_dashboard/main.tf
```diff
@@ -181,6 +181,38 @@ resource "datadog_dashboard" "application_metrics_dashboard" {
...
+        widget {
+          timeseries_definition {
+            title     = "Avg Time per Request"
+            live_span = var.widget_live_spans.apm
+            request {
+              display_type = "area"
+              query {
+                metric_query {
+                  name  = "query1"
+                  query = "sum:trace.${var.apm_primary_operation}.exec_time.by_service{service:${var.app}, $env} by {sublayer_service, sublayer_inferred}.rollup(sum).fill(zero)"
```
mianava, line 193: I think this has to become application:${var.app} and we'd mislabeled the other ones

terraform/modules/datadog_dashboard/main.tf
```diff
@@ -181,6 +181,38 @@ resource "datadog_dashboard" "application_metrics_dashboard" {
...
+                }
+              }
+              query {
+                metric_query {
+                  name  = "query2"
+                  query = "sum:trace.${var.apm_primary_operation}.hits{service:${var.app}, $env}.rollup(sum).fill(zero).as_count()"
```
mianava, line 199: same here

mianava (changes requested):

mianava (approved):
Approving as these will be non breaking and we can make modifications as necessary based on active data from multiple teams.

PR #526, opened 2026-07-20
BB2-4929 replace notify_no_data and no_data_timeframe with on_missing_data

annamontare-nava:
## 🎫 Ticket

BB2-4929

## 🛠 Changes

Switch to on_missing_data, which is more flexible and aligns with the Datadog UI normalization. 

Teams will have to update their configuration to use this change.

The monitors may need additional analysis to confirm the setting is correct when there is sparse data. This may vary between applications that have frequent requests vs infrequent requests. 

## ℹ️ Context

This seems to be the new way of doing things, and matches what's in the Datadog UI.

https://docs.datadoghq.com/monitors/guide/on_missing_data/

## 🧪 Validation

TODO


terraform/modules/datadog_monitors/lambda.tf
```diff
@@ -11,6 +11,9 @@ resource "datadog_monitor" "lambda_error_rate" {
     warning  = floor(var.monitor_config.lambda.error_rate_threshold * 0.75)
   }
 
+  # TODO this maybe doesn't make sense to be global across all the monitors for a technology
+  on_missing_data = var.monitor_config.lambda.on_missing_data
```
annamontare-nava, line 15: @mianava This monitor didn't have the notify_no_data options on it, was that intentional?

annamontare-nava: Same for some of the rds ones.

PR #534, opened 2026-07-22
BCDA-10269: update assertions schema +  support targetjsonpath

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10269

## 🛠 Changes

- Updated `assertions` schema to make `target` optional and added `targetjsonpath` object structure for JSONPath assertions.
- Added a dynamic `targetjsonpath` block inside `assertion` to support `validatesJSONPath` checks.

## ℹ️ Context

For BCDA, we are validating whether the response object from the _health endpoint has `ok` status for all dependencies. This PR extends the functionality of the CDAP DD synthetics module.


jasonvinson (approved):
LGTM!

PR #541, opened 2026-07-24
BB2-4929 Replace shadow_mode with draft_status everywhere

annamontare-nava:
## 🎫 Ticket

BB2-4929

## 🛠 Changes

Replace shadow_mode with the Datadog builtin draft_status in both shared and custom monitors.

## ℹ️ Context

shadow_mode and draft_status do approximately the same thing, and draft_status is a built in Datadog concept.

Adding a per-monitor draft_status for custom monitors allows teams to develop custom monitors one at a time without changing the global shadow_mode for that environment.

## 🧪 Validation

Tested on Blue Button.


mianava (changes requested):
If adding a draft mode, it might be appropriate to remove shadow_mode entirely

annamontare-nava:
> If adding a draft mode, it might be appropriate to remove shadow_mode entirely

That might make sense, is there anything shadow mode gives us that draft doesn't? I know that the drafts are nice because they show up more clearly in the Datadog UI, though they are deleted if not updated or published in 6 months.

mianava:
> > If adding a draft mode, it might be appropriate to remove shadow_mode entirely
> 
> That might make sense, is there anything shadow mode gives us that draft doesn't? I know that the drafts are nice because they show up more clearly in the Datadog UI, though they are deleted if not updated or published in 6 months.

I don't think so, @annamontare-nava, I think it would be safe to remove fully. The only intent was to not notify before we had settled any monitors and notification pipeline.

annamontare-nava:
@mianava I replaced shadow_mode with draft_status, but if you'd like me to just remove the option on the builtin monitors entirely, I can do that too.

mianava (approved):

PR #542, opened 2026-07-24
BCDA-10269: combine locations, eval min duration, add defaults

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10269

## 🛠 Changes

- Added optional `min_location_failed`, `min_failure_duration` fields to the `tests` variable schema
- Provided defaults in `variables.tf` so caller inputs remain optional

## ℹ️ Context

To prevent paging on single failures, BCDA requires a configurable minimum failure duration threshold before alerts trigger.

### DIFF
--- a/terraform/modules/datadog_synthetics/main.tf
+++ b/terraform/modules/datadog_synthetics/main.tf
@@ -33,11 +33,14 @@ resource "datadog_synthetics_test" "this" {
     }
   }
 
-  locations = each.value.use_private_location ? [local.private_location_id] : local.non_private_location_ids
+  locations = distinct(concat(
+    each.value.use_private_location && local.private_location_id != null ? [local.private_location_id] : (!each.value.use_private_location ? local.non_private_location_ids : []),
+    each.value.locations != null ? each.value.locations : []
+  ))
 
   lifecycle {
     precondition {
-      condition     = !each.value.use_private_location || local.private_location_id != null
+      condition     = !each.value.use_private_location || local.private_location_id != null || (each.value.locations != null && length(each.value.locations) > 0)
       error_message = "No Datadog private location found with prefix '${local.location_prefix}'. Verify the private location agent is registered for this environment."
     }
   }
@@ -45,7 +48,8 @@ resource "datadog_synthetics_test" "this" {
   options_list {
     tick_every           = each.value.tick_every
     monitor_name         = "[${upper(var.env)}] [${var.app}] Synthetics — ${each.value.name}"
-    min_failure_duration = var.min_failure_duration
+    min_failure_duration = try(coalesce(each.value.min_failure_duration, var.min_failure_duration), null)
+    min_location_failed  = each.value.min_location_failed
   }
 
   tags = concat(local.base_tags, each.value.tags)
--- a/terraform/modules/datadog_synthetics/main.tf
+++ b/terraform/modules/datadog_synthetics/main.tf
@@ -33,14 +33,11 @@ resource "datadog_synthetics_test" "this" {
     }
   }
 
-  locations = distinct(concat(
-    each.value.use_private_location && local.private_location_id != null ? [local.private_location_id] : (!each.value.use_private_location ? local.non_private_location_ids : []),
-    each.value.locations != null ? each.value.locations : []
-  ))
+  locations = each.value.use_private_location ? [local.private_location_id] : local.non_private_location_ids
 
   lifecycle {
     precondition {
-      condition     = !each.value.use_private_location || local.private_location_id != null || (each.value.locations != null && length(each.value.locations) > 0)
+      condition     = !each.value.use_private_location || local.private_location_id != null
       error_message = "No Datadog private location found with prefix '${local.location_prefix}'. Verify the private location agent is registered for this environment."
     }
   }
@@ -49,7 +46,6 @@ resource "datadog_synthetics_test" "this" {
     tick_every           = each.value.tick_every
     monitor_name         = "[${upper(var.env)}] [${var.app}] Synthetics — ${each.value.name}"
     min_failure_duration = try(coalesce(each.value.min_failure_duration, var.min_failure_duration), null)
-    min_location_failed  = each.value.min_location_failed
   }
 
   tags = concat(local.base_tags, each.value.tags)
--- a/terraform/modules/datadog_synthetics/variables.tf
+++ b/terraform/modules/datadog_synthetics/variables.tf
@@ -17,11 +17,13 @@ variable "shadow_mode" {
 variable "notify" {
   description = "Notify string from the monitors module."
   type        = string
+  default     = ""
 }
 
 variable "min_failure_duration" {
   description = "Minimum failure time to trigger alert in seconds. Should be set to the corresponding value from the monitor config passed to the monitors module."
   type        = number
+  default     = null
 }
 
 variable "tests" {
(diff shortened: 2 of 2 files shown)


terraform/modules/datadog_synthetics/variables.tf
```diff
@@ -65,8 +67,11 @@ variable "tests" {
       }))
     }))
 
-    tick_every = optional(number, 60)
-    tags       = optional(list(string), [])
+    tick_every           = optional(number, 60)
```
carlpartridge, line 68: This uses _seconds_ I imagine?

terraform/modules/datadog_synthetics/variables.tf
```diff
@@ -17,11 +17,13 @@ variable "shadow_mode" {
 variable "notify" {
   description = "Notify string from the monitors module."
   type        = string
+  default     = ""
```
annamontare-nava: I guess I'm wondering why you'd want to not pass in the notify string?

bhagatparwinder: You can always pass it. This just gives a default value if you do not want to pass it (making it optional).

annamontare-nava: I meant why do you want it to be optional?

terraform/modules/datadog_synthetics/variables.tf
```diff
@@ -65,8 +67,11 @@ variable "tests" {
       }))
     }))
 
-    tick_every = optional(number, 60)
-    tags       = optional(list(string), [])
+    tick_every           = optional(number, 60)
+    min_failure_duration = optional(number)
+    min_location_failed  = optional(number)
+    locations            = optional(list(string))
```
carlpartridge: So I see that datadog offers a number of American public locations and also for private locations.  I could see at some point us verifying for each?  In the meantime it seems like DD uses the min_failure_duration to allow for retries and corrections?

terraform/modules/datadog_synthetics/variables.tf
bhagatparwinder: If you want monitors but don't want alerting, for example, you create a dashboard in DD for the monitors. I'm more than happy to remove it and make it required.

terraform/modules/datadog_synthetics/variables.tf
bhagatparwinder: Correct. This was here, and I didn't update the default.

terraform/modules/datadog_synthetics/variables.tf
bhagatparwinder: DD provides three locations at the time (AWSGov public location and two CDAP private locations). We are testing with all provided (see bcda-app repo to check the location I am providing).

No, min_failure_duration is just duration of the failure. DD has a different flag for retries. It is an object with two values. Example below:

```js
{
        tick_every = 1800
        min_failure_duration = 3600
        min_location_failed = 2
        http_version = "any"
        retry {
            count = 0
            interval = 300
        }
        monitor_options {
            notification_preset_name = "show_all"
        }
    }
```

terraform/modules/datadog_synthetics/variables.tf
annamontare-nava: Oh okay, that makes sense!

Some context on why I didn't specify a default when I added this variable: I was intending for the notify string to be the same in the synthetics as in the monitors module, and so I did not set a default so that if you forgot to connect the two modules, you'd get an error.

terraform/modules/datadog_synthetics/variables.tf
```diff
@@ -17,11 +17,13 @@ variable "shadow_mode" {
 variable "notify" {
   description = "Notify string from the monitors module."
   type        = string
+  default     = ""
 }
 
 variable "min_failure_duration" {
   description = "Minimum failure time to trigger alert in seconds. Should be set to the corresponding value from the monitor config passed to the monitors module."
   type        = number
+  default     = null
```
annamontare-nava: Same sort of question here, why do you want it to be optional?

annamontare-nava: I made it required because I expected it to be set to the value read from the config file, and I wanted an error if someone forgot that connection.

bhagatparwinder: This was not to force the caller to pass it if they don't have a specific requirement. For example, we at BCDA wanted 60 minutes. If someone doesn't pass it, DD uses the default. However, I have removed the defaults from both `notify` and `min_failure_duration` in the latest commit.

carlpartridge (dismissed):

mianava (dismissed):
We were not intending on supporting requests from a non-private location, one hosted by Datadog, as this would mean, especially on API endpoints, enabling datadog ingress instead of supporting only in-network traffic for these calls. I think on a static public site, the public locations option would make sense, but if I'm understanding the implications of location setup in the PR, we should instead favor a secondary region private location and err towards having more private network configurations.

bhagatparwinder:
> We were not intending on supporting requests from a non-private location, one hosted by Datadog, as this would mean, especially on API endpoints, enabling datadog ingress instead of supporting only in-network traffic for these calls. I think on a static public site, the public locations option would make sense, but if I'm understanding the implications of location setup in the PR, we should instead favor a secondary region private location and err towards having more private network configurations.

This isn't meant for private endpoints. It is for public resources (such as the static site or the _sandbox environment API_). For any private endpoints, yes, it makes sense to test them from a secondary private location.

We are trying not to page folks in the middle of the night if there is a transient failure (due to resource access issues or location issues). So, with the help of this PR (and the accompanying https://github.com/CMSgov/bcda-app/pull/1449), we are going from 1 failure at 1 location to 2 failures at 2 locations.

carlpartridge (dismissed):

terraform/modules/datadog_synthetics/variables.tf
```diff
@@ -65,8 +65,9 @@ variable "tests" {
       }))
     }))
 
-    tick_every = optional(number, 60)
-    tags       = optional(list(string), [])
+    tick_every           = optional(number, 60)
+    min_failure_duration = optional(number)
```
mianava: I am cringing a bit at this comment: 
can this switch to option(number, null) ? to align with standards

terraform/modules/datadog_synthetics/main.tf
```diff
@@ -45,7 +45,7 @@ resource "datadog_synthetics_test" "this" {
   options_list {
     tick_every           = each.value.tick_every
     monitor_name         = "[${upper(var.env)}] [${var.app}] Synthetics — ${each.value.name}"
-    min_failure_duration = var.min_failure_duration
+    min_failure_duration = try(coalesce(each.value.min_failure_duration, var.min_failure_duration), null)
```
mianava, line 48: With the null in the definition, would this be possible without the try? as var.min_failure_duration will fallback to null?

mianava, line 48:
```suggestion
    min_failure_duration = coalesce(each.value.min_failure_duration, var.min_failure_duration)
```

terraform/modules/datadog_synthetics/variables.tf
mianava:
```suggestion
    min_failure_duration = optional(number, null)
```

mianava (changes requested):

terraform/modules/datadog_synthetics/variables.tf
bhagatparwinder: Fair, updated!

terraform/modules/datadog_synthetics/main.tf
bhagatparwinder: This issue is how `coalesce` works. It requires one non-null value in the list; otherwise, it returns an error.

It could, however, be written as a ternary operator.

```
min_failure_duration = each.value.min_failure_duration != null ? each.value.min_failure_duration : var.min_failure_duration
```

mianava (approved):

PR #543, opened 2026-07-24
BB2-4929 remove ALERT_THRESHOLD from webhooks

annamontare-nava:
## 🎫 Ticket

BB2-4929

## 🛠 Changes

Remove the non-existent `ALERT_THRESHOLD` variable from the Datadog webhooks.

This will also need to be removed from the Slack workflows, and maybe from VictorOps?

## ℹ️ Context

Alert threshold is not one of the parameters supported by webhooks in
Datadog, so remove it.

https://docs.datadoghq.com/integrations/webhooks/#variables

## 🧪 Validation

I think someone on CDAP will need to validate this.

### DIFF
--- a/terraform/services/505-datadog-webhooks/main.tf
+++ b/terraform/services/505-datadog-webhooks/main.tf
@@ -26,7 +26,6 @@ resource "datadog_webhook" "slack_channels" {
     status     = "$ALERT_STATUS"
     priority   = "$PRIORITY"
     metric     = "$ALERT_METRIC"
-    threshold  = "$ALERT_THRESHOLD"
     message    = "$TEXT_ONLY_MSG"
     host       = "$HOSTNAME"
     tags       = "$TAGS"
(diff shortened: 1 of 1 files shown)


terraform/services/505-datadog-webhooks/main.tf
```diff
@@ -62,7 +61,6 @@ resource "datadog_webhook" "victorops_endpoints" {
     monitoring_tool     = "Datadog"        # Identifies the source tool
     host_name           = "$HOSTNAME"      # Affected host
     metric_name         = "$ALERT_METRIC"  # The metric that triggered the alert
-    alert_threshold     = "$ALERT_THRESHOLD"
```
mianava, line 65: Ah, based on https://docs.datadoghq.com/integrations/webhooks/#usage , yep. As far as I can tell, this value was not used to configure the slack webhooks.

mianava (approved):

terraform/services/505-datadog-webhooks/main.tf
annamontare-nava: It was in our slack webhook for a while, and just showed up as $ALERT_THRESHOLD in slack.

PR #546, opened 2026-07-28
BB2-4929 Add variable to turn on and off replica lag monitor

annamontare-nava:
## 🎫 Ticket

BB2-4929

## 🛠 Changes

Add the option to turn off the Aurora replica lag monitor.

## ℹ️ Context

Not an environments for a team might have replicas, so this monitor may
not have anything to look at

## 🧪 Validation

Tested on Blue Button


mianava (approved):
Makes sense as backup can be handled otherwise

PR #548, opened 2026-07-29
BB2-4929 Rename the env template variable to environment

annamontare-nava:
## 🎫 Ticket

BB2-4929

## 🛠 Changes

Rename the template variable `env` in the shared dashboard to `environment`.

## ℹ️ Context

The tag it actually filters by is `environment`, so naming it the same
makes things clearer.

## 🧪 Validation

Tested on Blue Button.


mianava (approved):
This consistency seems great!

PR #551, opened 2026-07-30
PLT-1719 Add an ALB module that supports external or internal use.

mianava:
## 🎫 Ticket

PLT-1719

## 🛠 Changes
Merge 1

Introduces a new OpenTofu module for provisioning an AWS ALB with opinionated/platform-aware defaults. This module can be consumed by terraservices in application repos that need a managed ALB with listener and security group built in. This ALB can be either internal or external, though internal use will be minimized as most endpoints can connect via service connect. 

## 🧪 Validation

These changes were tested in a subsequent PR that creates a full ecs stack behind an internal load balancer, and another one that does the same behind an external load balancer. 
This module is not currently used and will not break any downstream resoruces or prevent their provisioning if merged.


terraform/modules/alb/security_group_rules.tf
```diff
@@ -0,0 +1,56 @@
...
+# CMS security tools ingress — always on when module manages the SG
+resource "aws_vpc_security_group_ingress_rule" "https_security_tools" {
+  count                        = local.managed_sg ? 1 : 0
+  security_group_id            = aws_security_group.alb[0].id
+  referenced_security_group_id = var.platform.security_groups["cmscloud-security-tools"].id
+  from_port                    = 443
+  to_port                      = 443
+  ip_protocol                  = "tcp"
+  description                  = "Allow HTTPS from CMS security scanning tools"
+}
```
mianava, line 56: @bhodges-navapbc These are the security group rules for review in the ALB context. This is what can be provided with basic flags into the ALB module. Could you help me confirm this is what is accurately needed for cmscloud-security-tools?

jasonvinson (dismissed):

jasonvinson (approved):

PR #553, opened 2026-07-30
PLT-1720 Build, test, and deploy a service without ALB, ACM, or service connect

mianava:
## 🎫 Ticket
PLT-1720 

## 🛠 Changes

Updates the service module to support a sidecar, specifically an mTLS sidecar, but not use it. This validates backwards compatibility with existing services that do not use mTLS sidecar or act as a target group for ALB. 
This PR also establishes a structure and deployment process for 
- docker hosted applications that are automatically tested and deployable via git workflow
- testing via sonarqube for Python 
- Makefile usage for deployment and local development 

## ℹ️ Context
These changes are made as 

## 🧪 Validation
This module implements a service that is testable through a Makefile and by provisioning the service, which is confirmed via the AWS GUI.  

The git workflow was also used to build and test from this branch, though on merge to main we will confirm the provisioning workflow occurs as expected.


terraform/modules/service/iam.tf
```diff
@@ -147,7 +150,7 @@ data "aws_iam_policy_document" "service_connect" {
       condition {
         test     = "StringLike"
         variable = "acm:DomainName"
-        values   = ["*.${var.service_connect_namespace.name}"]
+        values   = ["*.${var.platform.app}-${var.platform.env}.sc.internal.cms.gov"]
```
mianava, line 153: Security between clusters would be improved if we still referenced name.

jasonvinson (approved):

PR #559, opened 2026-08-06
PLT-1747 GHA Role Reduction

mianava:
## 🎫 Ticket

PLT-1747

## 🛠 Changes

- replaced the single aws_iam_role_policy inline policy on the ${var.app}-${var.env}-github-actions role with multiple aws_iam_policy managed policies, grouped by domain
- removed the aws_iam_instance_profile and the ec2.amazonaws.com assume role statement as the role is no longer used for Packer or direct EC2 instance launching
- removed codebuild permissions from the shared policy and consolidated into a new cdap-only policy alongside secretsmanager, since only CDAP manages CodeBuild infrastructure and uses Secrets Manager
- removed firehose permissions as no app currently manages firehose through opentofu (or at all) 
- Collapsed verbose Describe*, List*, and Get* action enumerations into glob patterns where safe to do so to reduce length of all policies 
- removed direct EC2 instance launching actions (RunInstances, CreateImage, CreateKeyPair, DeleteKeyPair, ModifyImageAttribute) as teams no longer spin up EC2 instances directly
- consolidated app-specific KMS aliases (previously hardcoded as individual data blocks with count conditionals) into per-app, per-env YAML config files under config/${var.app}/${var.env}.yml, leveraging the existing additional_kms config pattern

- Fixed two bugs in the previous KMS data sources:
    aws_kms_alias.ab2d_tfstate_bucket was incorrectly gating on var.env == "ab2d" instead of var.app == "ab2d"
    aws_kms_alias.bcda_aco_creds had "dev, test" as a single string instead of ["dev", "test"]

## ℹ️ Context

Future Improvements

These improvements are flagged as FIXMEs but separate work must be completed first:

- SSM path scoping: Resource ARNs for SSM actions are currently "*". Tightening these to parameter/${var.app}/${var.env}/* requires first auditing and standardizing SSM parameter path conventions across all apps, with particular focus on CDAP-hosted shared tokens. 
- IAM resource scoping: IAM mutating actions are currently scoped to "*". These can be tightened to arn:aws:iam::*:role/${var.app}/${var.env}-* once IAM role/policy naming conventions are confirmed consistent across all apps and environments
- Verify use of account_env and account_env_old KMS keys 
- S3 action scoping: s3:PutBucket* and similar write actions can be scoped to known bucket name patterns once naming conventions are validated

## 🧪 Validation

This change will be applied team by team, environment by environment before merging:

- run tofu plan for each app/env combination and verify the plan shows only the expected destroy/create changes 
- run available tofu apply automation for each app/env and confirm successful apply
- merge only after all apps and environments have been successfully validated


jasonvinson (dismissed):

jasonvinson (dismissed):

terraform/services/005-github-actions-role/diff.sh
```diff
@@ -0,0 +1,129 @@
+#!/bin/zsh
```
gfreeman-navapbc: neat script

gfreeman-navapbc (dismissed):

gfreeman-navapbc (approved):

